PR #12, opened 2016-09-29
Latest integration w/Joel's work

amardesi:


martinhaye:
Great work Andy. 

The only thing I see missing is the customizations Amy made to the Search object (which is now a component). Looks like she adapted Joel's code to use react-router-form and <Form> element (upper-case F) instead of <form> (lower-case f). You'll know it's working if you type a search term into the search box, hit enter, and it goes to the search page.

amardesi:
I had made that change but must have accidentally overwritten that. Fixed now.

martinhaye:
Lovely!

PR #14, opened 2016-10-06
Integrating pdf.js. viewer into ItemPage.

martinhaye:
Andy, could you check what I've done to ItemPage and see if you're happy with it?


martinhaye:
Andy says it's fine. Going ahead.

PR #25, opened 2016-10-27
Various item page stuff

amardesi:


amardesi:
It might make sense to rename breadcrumb.rb to hierarchy.rb

PR #36, opened 2016-12-02
Refactoring database tables for widgets and pages

martinhaye:
Andy and Amy, I'd like your comments before I try to run this code. Does this match what we agreed to today?


amywieliczka:
Looks good to me.

PR #40, opened 2016-12-12
Merged Joel's most recent eschol-ui code.

martinhaye:
This was a real 3-way merge. Most things weren't too difficult, the exception being the split out Header and Nav components. I've stubbed that for now so it roughly works; Andy has offered to detangle.


martinhaye:
Just going to go ahead and merge this to the main repo. It works well enough that we can all continue dev, while Andy and Joel sort out the new header/nav components.

PR #41, opened 2016-12-16
Browse Pages

amardesi:
Departments are not selectable. Might make sense to wait on Joel's work to do final work on this.
Also need to bring in proper top navigation for Departments page.

Besides these departments issues, all other browse functionality is complete.


app/jsx/pages/BrowsePage.jsx
```diff
@@ -60,6 +64,90 @@ felis tellus mollis orci, sed start a journal nunc eget odio.
...
+      } else {
+        this.setState({children: this.props.depts.children})
+      }
+    }
+    ev.preventDefault()
+    ev.stopPropagation()
+  }
+  render() {
+    if (!this.state.children) this.state.children = []
+    return (
```
martinhaye: When I run this code, no matter what I click on (label or the little bullet) both do a child display toggle. The <Link> never activates for me. Is that on your roadmap?

app/jsx/pages/BrowsePage.jsx
```diff
@@ -60,6 +64,90 @@ felis tellus mollis orci, sed start a journal nunc eget odio.
...
+//    [{"id"=>"asia",
+//      "name"=>"UCLA Asia Institute",
+//      "children"=>[{"id"=>"ccs", "name"=>"UCLA Center for Chinese Studies"}]}, .. ], } .. ]
+class DeptTree extends React.Component {
+  constructor(props) {
+    super(props)
+    this.state = { depts: this.props.depts }
+    // this.onSelect = this.onSelect.bind(this)
+  }
+  onSelect(node) {
```
martinhaye: What's the goal of this onSelect code? It seems like it's trying to deselect the previously selected node, but if that's happening I'm not understanding the effect. Does "select" here mean "expand" in the UI? If so, it seems like this would make it so only one is expanded at a time, which I'm not seeing when I run it.

app/jsx/pages/BrowsePage.jsx
```diff
@@ -77,11 +165,15 @@ class BrowseJournals extends React.Component {
 //       :ancestor_unit=>["ucla", "ucsc"],
 //       :is_active=>true}
   getVisibleJournals(journals, isActive, campusID) {
-    return journals.map(function(j, i) {
-      return (j['is_active'] == isActive || isActive == "") &&
+    let foundOne = false
+    let r = journals.map(function(j, i) {
```
martinhaye: Just wondering whether there's any value to the server returning inactive journals at all. If the server filtered them out, they'd be unnecessary here.

Or maybe the thinking was, if the user is logged as an admin, then you'd show the inactive journals? That could make some sense.

amardesi: Wireframes indicate ability to choose archived items as well as actively publishing. http://ux.cdlib.org/jkratz/eScholarship/wireframes/eScholarship_2016-12-6/#g=1&p=browse_journals

app/jsx/pages/BrowsePage.jsx
amardesi: I was going to wait for Joel's work to complete before getting this to work. Does that seem appropriate to you?

app/jsx/pages/BrowsePage.jsx
amardesi: Martin, I did not think through the onSelect at all. We can remove it for now if you like.

martinhaye: Understood. If it's just temporary code, that's fine.

app/jsx/pages/BrowsePage.jsx
martinhaye: Yes, that seems appropriate.

PR #48, opened 2017-02-01
Big app-state refactor.

martinhaye:


martinhaye:
We reviewed the essence of this at today's codeAndTell.

PR #61, opened 2017-03-01
Joel's latest changes

amardesi:


app/jsx/components/BreadcrumbComp.jsx
```diff
@@ -5,21 +5,24 @@ import { Link } from 'react-router'
...
-      if (node.id == 'root') {
-        return (
-          <Link className={(i==lastN) ? "c-breadcrumb-link--active": null} to="/" key={i}>{node.name}</Link>
-        )
-      }
-      return (
-        <Link className={(i==lastN) ? "c-breadcrumb-link--active": null} to={"/unit/" + node.id} key={i}>{node.name}</Link>
-      )
-    });
+    let missingNavStyle = this.props.array ? null : { margin: '20px 0px 20px' },
```
amardesi: Items with no unit parent will no display a NavBarComp. Yet, the NavBarComp includes a bottom margin of 20px. This situation also begs whether we even need to display a breadcrumb. Either way, thing look ugly without me hardcoding this style into the breadcrumb for this use case and we should probably need to bring UX into the discussion eventually.

app/jsx/components/AuthorListComp.jsx
```diff
@@ -1,46 +1,33 @@
...
     return (
       <div className="c-authorlist">
-        <time className="c-authorlist__year">{year && year[0]}</time>
+        {year && <time className="c-authorlist__year">{year[0]}</time> }
         { p.authors && 
             <ul className="c-authorlist__list">      
-              {authorlist} {this.renderExpand(expand)}
+              {authorlist}
+              <li>
+                <Link to="#" className="c-authorlist__list-more-link" onClick={(e)=>this.handleClick(e, 4)}>et al.</Link>
```
amardesi: This et al link is not working for me for some reason. But I'm going to wait to fix until Joel resolves the other bug going on with the et al link showing up no matter how many lines of the authorlist display.

app/jsx/components/BreadcrumbComp.jsx
martinhaye: I like the workaround here. I'd suggest adding your comment here as a comment in the code as well, but that won't stop me from merging this PR.

PR #71, opened 2017-03-28
Touchups to Item page and latest Joel work

amardesi:


app/jsx/components/Search2Comp.jsx
```diff
@@ -22,15 +22,15 @@ class SearchControls extends React.Component {
     if (p.unitID) {
       searchUnit = (
         <span>
-          <input type="radio" id="c-search2__refine-campus" name="searchType" value={p.unitID} onClick={this.props.handleRadioSelect} />
+          <input type="radio" id="c-search2__refine-campus" name="searchType" value={p.unitID} />
```
martinhaye: Joel did some recent work on focus and blur involving the search box and its radio buttons, but that doesn't seem to have been carried over. In addition, styling of the search box is subtly different between jschol vs Joel's design. Can you look into what we might have missed merging here?

PR #77, opened 2017-04-21
Item view update and some other cleanup

amardesi:


app/jsx/components/JumpComp.jsx
```diff
@@ -15,16 +15,18 @@ class JumpComp extends React.Component {
       <div className="c-jump">
         <Link id="c-jump__label" to="#" onClick={(e)=>this.handleClick(e, 1)}>Article</Link>
         <ul className="c-jump__tree" aria-labelledby="c-jump__label">
+       { this.props.attrs.abstract &&
           <li><Link to="#">Abstract</Link></li>
+       }
```
martinhaye: It took me a while, but I get what you're doing with the indentation here. Trying to preserve mergeability with Joel's code right? Clever.

app/jsx/pages/BrowsePage.jsx
```diff
@@ -28,7 +28,6 @@ class BrowsePage extends PageBase
   renderData(data) {
     return (
       <div>
-        <a href="#maincontent" className="c-skipnav">Skip to main content</a>
```
martinhaye: Why are we taking out skip links? These are for keyboard users.

martinhaye: Nevermind, I get it... there's a SkipNavComp now. Coolness!

PR #79, opened 2017-04-26
Tab Hash in Item View; Backout pdfjs-embed bower change; Item page logic

amardesi:
Martin,  I had to add one more line to ItemPage: line 39
`tabName = (tabName.startsWith("article") ? "main" : tabName)`
Otherwise it was setting tabName = article_main


amardesi:
Did a bit of refactoring, and also corrected this problem: Proper Tab was not rendering upon first render of a URL including a hash.

Still need to correct this issue which occurs when one returns to the main tab:
"Error: The overlay is already registered."
I see pdfjs is intializing every time you come back to this tab. Seems wrong to me. Would love your help when you're free.

app/jsx/pages/ItemPage.jsx
```diff
@@ -11,15 +11,45 @@ import TabsComp from '../components/TabsComp.jsx'
 import JumpComp from '../components/JumpComp.jsx'
 import FooterComp from '../components/FooterComp.jsx'
 
+const anchors = ['main', 'supplemental', 'metrics', 'author',
+             'article_abstract', 'article_main', 'article_references']
+
 class ItemPage extends PageBase {
-  state = { currentTab: 1 }
+  static propTypes = {
+    currentTab: React.PropTypes.oneOf(anchors)
```
martinhaye: Ooh nice!

app/jsx/pages/ItemPage.jsx
```diff
@@ -11,15 +11,45 @@ import TabsComp from '../components/TabsComp.jsx'
...
 
-  changeTab = tab_id => { this.setState({currentTab: tab_id}) }
+  tabNameFromHash() {
+    return location.hash.toLowerCase().replace(/^#/, "")
+  }
+
+  // currentTab should be 'main' whenever hash starts with 'article_" (or is empty)
+  articleHashHandler = h => {
+    return ((h.startsWith("article") || h=='') ? "main" : h)
+  }
```
martinhaye: For future ref, I think this can be simplified to a one-liner:
`articleHashHandler = h => ((h.startsWith("article") || h=='') ? "main" : h)`

bower.json
```diff
@@ -13,6 +13,7 @@
     "normalize-scss": "^3.0.3",
     "flickity": "^2.0.5",
     "babel-polyfill": "^0.0.1",
+    "pdfjs-embed": "eScholarship/pdfjs-embed",
```
martinhaye: Is there a way we can specify the particular changeset id to bower (like we do in package.json) to force it to pick up a new version when we change CSS or image resources in pdfjs-embed?

amardesi: Yes! same way as package.json, by adding #<branch>. Change this and will be adding it as next commit momentarily.

PR #81, opened 2017-05-08
Browse pages proper URL; Global (root) Nav established for browse and static pages that need them

amardesi:


app/jsx/pages/BrowsePage.jsx
```diff
@@ -7,7 +7,7 @@ import PageBase from './PageBase.jsx'
 import Header1Comp from '../components/Header1Comp.jsx'
 import Header2Comp from '../components/Header2Comp.jsx'
 import Subheader2Comp from '../components/Subheader2Comp.jsx'
-import Nav1Comp from '../components/Nav1Comp.jsx'
+import NavComp from '../components/NavComp.jsx'
```
amardesi: Will plan on removing all references to Nav1Comp and Nav2Comp. This is just a start.

app/server.rb
```diff
@@ -453,22 +461,27 @@ def getChildDepts(unit)
...
+    authors = ItemAuthors.filter(:item_id => id).order(:ordering).
+                 map(:attrs).collect{ |h| JSON.parse(h)}
+    citation = getCitation(shortArk, authors, attrs)
     begin
       body = {
         :id => shortArk,
+        :citation => citation,
+        :title => citation[:title],
+        # ToDo: Normalize author attributes across all components (i.e. 'family' vs. 'lname')
+        :authors => authors,
```
amardesi: If we do in fact use the CSL-JSON schema, author needs to use different attribute names than what we have in our DB.  Sticking this here for now so authorlist doesn't break. But in the future it would be best to have just one authorlist sitting in citation.

amardesi:
I wasn't sure how to avoid including my latest work in this pull request. Does it matter that it's just getting appended into this pull request?

app/jsx/components/TabAuthorComp.jsx
```diff
@@ -34,34 +54,36 @@ class TabAuthorComp extends React.Component {
...
+        {p.attrs['orig_citation'] ?
           <dl className="c-descriptionlist">
-            {/* ToDo: Bring in all citations styles */}
-            <dt><a href="">APA</a></dt>
-            <dd>
-              {p.attrs['orig_citation']}
-            </dd>
+            <dt></dt><dd>{p.attrs['orig_citation']}&nbsp;&nbsp;{retrieved_suffix}</dd>
+          </dl>
+          :
```
amardesi: Am I correct in assuming that the orig_citation is the custom citation?

app/citation.rb
```diff
@@ -0,0 +1,44 @@
...
+def getAuthors(a)
+  if !a
+    return nil
+  else
+    a_mapped = []
+    # ToDo: map middle name when it's present: {fname"+"middle" => "given"}
+    #       and handle for when it's not
+    mappings = {"name" => "literal", "lname" => "family", "fname" => "given"} 
+    a.each do |x|
+      a_mapped << x.map {|k, v| [mappings[k]||k, v]}.to_h
```
martinhaye: Fancy!

app/jsx/App.jsx
```diff
@@ -35,8 +35,9 @@ class App extends React.Component
 const routes = (
   <Route path="/" component={App}>
     <IndexRoute component={HomePage} />
-    <Route path="/browse/:browse_type" component={BrowsePage} />
-    <Route path="/browse/depts/:campusID" component={BrowsePage} />
+    <Route path="/campuses" component={BrowsePage} />
+    <Route path="/journals" component={BrowsePage} />
```
martinhaye: Yay, this is a nice change.

app/server.rb
martinhaye: I don't know that I want to modify our database to match the CDL-JSON schema; it would seem like overkill to me. And actually, you could put the translation code on the React side of things.

PR #83, opened 2017-05-23
Browse journals and Item Supplemental files

amardesi:


app/jsx/components/ScholWorksComp.jsx
```diff
@@ -40,12 +40,8 @@ class ScholWorksComp extends React.Component {
 
     if (this.props.result.authors) {
       var authorList = this.props.result.authors.map(function(author, i, a) {
-        if (i === a.length-1) {
```
amardesi: dotdotdot nullifies the need for this logic

martinhaye: Actually there's still a need to put semicolons between the authors.

PR #85, opened 2017-05-25
Browse campus journals 2nd pass; Merged Joel's latest eschol-ui

amardesi:


amardesi:
Martin, if you could accept this PR then all the browse pages will look pretty god for our meeting tomorrow.

PR #93, opened 2017-06-09
Campus Landing page 1st pass

amardesi:


app/jsx/pages/BrowsePage.jsx
```diff
@@ -44,13 +43,13 @@ class BrowsePage extends PageBase
         // Campus-specific browse page
         <div>
           <Header2Comp type="campus" unitID={data.campusID} />
-          <Subheader2Comp unit={data.unit}
-                          campusID={data.campusID}
-                          campusName={data.campusName}
+          <Subheader2Comp unit={data.unit} logo={data.header.logo}
```
amardesi: I see Joel has wrapped all subheaders into one component now.....I'm working on headers next week.

app/jsx/pages/BrowsePage.jsx
```diff
@@ -44,13 +43,13 @@ class BrowsePage extends PageBase
...
+                          campusID={data.header.campusID}
+                          campusName={data.header.campusName}
                           campuses={data.campuses} />
           <div className="c-navbar">
             {/* ToDo: Properly call header.nav_bar for unit type="campus" */}
-            <NavComp data={[{name: 'Open Access Policies', slug: ''}, {name: 'Journals', slug: '/' + data.campusID + '/journals'}, {name: 'Academic Units', slug: '/' + data.campusID + '/units'}]} />
+            <NavComp data={[{name: 'Open Access Policies', url: ''}, {name: 'Journals', url: '/' + data.header.campusID + '/journals'}, {name: 'Academic Units', url: '/' + data.header.campusID + '/units'}]} />
```
amardesi: FYI, wireframes indicate that Campus level NavBar will always be set to: "Open Access Policies | Journals | Academic Units". DB currently does not reflect this, so I'm just temporarily displaying them here so links work.

app/jsx/pages/BrowsePage.jsx
```diff
@@ -105,11 +106,15 @@ class AllCampuses extends React.Component {
 class AllJournals extends React.Component {
   constructor(props) {
     super(props)
-    this.state = {campusID: props.campusID,
+    this.state = {campusID: "",
                   isActive: props.isActive}
     this.changeCampus = this.changeCampus.bind(this)
   }
 
+  componentWillUnmount() {
```
amardesi: I noticed the input selector wasn't being reset when I clicked back to this page. Ideally it would be nice to preserve all state on this page, I wasn't able to get that working. For now, if you click back to this page, all journals display (and input selector should be set to "All Campuses").

app/jsx/pages/BrowsePage.jsx
martinhaye: Hmm, I'd like to verify with Justin and John that admins will not be able to add anything to the campus nav bar.

app/jsx/pages/BrowsePage.jsx
martinhaye: There are a couple ways to get it to retain its setting. I think the clearest is to add it as a parameter in the URL. I've not tried the more secretive ways to saving state.

app/jsx/pages/CampusPage.jsx
```diff
@@ -0,0 +1,105 @@
...
+
+  dashUrlList = {
+    'ucb': 'https://dash.berkeley.edu/stash',
+    'uci': 'https://dash.lib.uci.edu/stash',
+    'ucm': 'https://dash.ucmerced.edu/stash',
+    'ucop': 'https://dash.ucop.edu/stash',
+    'ucr': 'https://dash.ucr.edu/stash',
+    'ucsc': 'https://dash.library.ucsc.edu/stash',
+    'ucsf': 'https://datashare.ucsf.edu/stash',
+  }
```
martinhaye: Eventually the dash URL should be an attr on the entity that's configured on the campus CMS page. But this is fine for now.

PR #98, opened 2017-06-15
Headers fully integrated, including Campus Page

amardesi:


amardesi:
I added social links to ucla and uclalaw to test these in the navbar. Test is complete. But they'll get wiped out during your next conversion.

app/jsx/pages/SearchPage.jsx
```diff
@@ -240,14 +241,11 @@ class PubYear extends React.Component {
 //   }
 // }
 class FacetFieldset extends React.Component {
-  handleChange = this.handleChange.bind(this);
-  pubDateChange = this.pubDateChange.bind(this);
```
martinhaye: Glad to see you're on board with cleaning these up too.

app/jsx/pages/SearchPage.jsx
```diff
@@ -402,17 +400,25 @@ class FacetForm extends React.Component {
     return true;
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (!_.isEqual(this.props.query, nextProps.query))
+      this.setState({query: nextProps.query})
+  }
```
martinhaye: Ah! Good find.

app/jsx/pages/SearchPage.jsx
```diff
@@ -455,12 +462,17 @@ class SearchPage extends PageBase {
                 <h2 className="o-columnbox1__heading">
                   Scholarly Works ({data.count + " results" + (data.count > 10000 ? ", showing first 10000" : "")})</h2>
               </header>
+            {(data.count > 10) &&
```
martinhaye: Good point that this should be checked! However, instead of a constant 10, shouldn't it check against the currently selected number of results per page?

PR #101, opened 2017-06-19
Search Page work

amardesi:


app/jsx/components/FilterComp.jsx
```diff
@@ -54,9 +54,9 @@ class FilterComp extends React.Component {
       <div className="c-filter">
         <h2 className="c-filter__heading">{searchString}</h2>
         <input type="hidden" name="q" value={this.props.query.q} />
-        <div>Results: {Math.ceil(this.props.count/10)} pages, {this.props.count} works</div>
+        <div>Results: {Math.ceil(this.props.count/this.props.query.rows)} pages, {this.props.count} works</div>
       {activeFilters &&
-        <FiltersShowHide activeFilters={activeFilters} handler={this.props.handler} />
+        <FiltersShowHide activeFilters={activeFilters} handler={this.props.handler} clearAll={this.clearAll} />
```
amardesi: New sub-component that will - when implemented - only show search filters when user requests that.

app/jsx/pages/SearchPage.jsx
```diff
@@ -299,34 +302,50 @@ class FacetFieldset extends React.Component {
...
+            }
+          </div>
+        </details>
+      )
     } else {
-      //pub_year
-      facetItemNodes = (
-        <PubYear data={this.props.data} query={this.props.query} handler={this.pubDateChange} />
+      return (
+        <ModalComp isOpen
```
amardesi: Martin I'd like your help here. The modal is currently only displaying for a millisecond. I'm suspecting this is because the modal should be called from a higher level component?

app/unitPages.rb
```diff
@@ -174,6 +170,7 @@ def getIssue(id)
 
 
 def unitSearch(params, unit)
+  # ToDo: remove this section now that unit search display looks just like global search
```
amardesi: It would be really nice to get rid of duplicative unitPages.unitSearch but I see series landing page needs it for sorting. Hrmmmm.

app/jsx/components/PaginationComp.jsx
```diff
@@ -96,8 +98,10 @@ class PaginationComp extends React.Component {
       return (
         <div className="c-pagination">
           <a href="" className="c-pagination__prevnext" onClick={this.previous}>Previous</a>
-          <a href="" className="c-pagination__item" onClick={this.first}>1</a>
-          <span className="c-pagination__ellipses">&hellip;</span>
+          { (pages > 3) && 
+            [<a key="0" href="" className="c-pagination__item" onClick={this.first}>1</a>,
+             <span key="1" className="c-pagination__ellipses">&hellip;</span>]
+          }
```
amardesi: If you need an example of search results that give you less than 4 pages worth: "crepuscular sheep"

app/jsx/pages/SearchPage.jsx
martinhaye: In the `<a href="" onClick=...>` line, you need to call e.preventDefault() in addition to the state work you're already doing. Otherwise, the browser goes ahead and treats the <a> as an actual hyperlink to url "", causing the page to reload. I tried it in my copy of your code and that resolved the problem.

app/jsx/components/PaginationComp.jsx
martinhaye: Cool query! This should go into Lisa's test plan... maybe send it to her.

I've never heard the word crepuscular used outside of the context of ocean creatures, but of course it makes perfect sense.

app/unitPages.rb
martinhaye: Well, we can refactor at some point. For now I'd say at least update the comment to make clear what the code is currently used for.

app/jsx/pages/SearchPage.jsx
amardesi: Duh, thanks

PR #108, opened 2017-06-29
Search2Comp fix

amardesi:


app/server.rb
```diff
@@ -591,17 +591,14 @@ def getChildDepts(unit)
                'campuses', 'departments', 'journals', 'disciplines', 'rights']
   params = CGI::parse(request.query_string)
   searchType = params["searchType"][0]
-  # ToDo:Could there be more than one searchType? 
+  # Perform global search when searchType is assigned 'eScholarship'
+  # otherwise: 'searchType' will be assigned the unit ID - and then 'searchUnitType' specifies type of unit.
```
martinhaye: I take it these special fields are being made so that, in future, we can add unit branding to the search page. Yes?

amardesi: While we only need one parameter to say whether it's a global search ...searching by unit needs two parameters: the unitID and what type it is (oru, campus, etc)
But yes, unit branding will rely on the presence of these unit level parameters

PR #109, opened 2017-06-30
Campus now under Unit page; NotYetLink

amardesi:
Made a little progress on campus landing page. Moving on to other landing pages before I work any more on campus-specific stuff.


app/jsx/components/NotYetLink.jsx
```diff
@@ -0,0 +1,33 @@
...
+
+  closeModal = e => {
+    this.setState({modalOpen:false})
+  }
+
+  content = (<div>Ahem...This page or functionality is not yet working.<br/>We appreciate your patience.</div>)
+
+  render() {
+    return (
+      <div id="notYetLinkModalBase">
```
martinhaye: I think this id and the parentSelector below aren't necessary for this component, and will in fact break things if there are multiple NotYetLinks on a page. The only reason we messed around with those for the Department modal was because we needed it to be nested within the form. I think react-modal's default behavior (that is, without parentSelector) will probably work fine for NotYetLink.

amardesi: Ah right, I'll fix, thanks.

PR #164, opened 2017-08-10
PDFjs loadingBar (loading status)

amardesi:
There's still a box-shadow around the toolbarContainer that I couldn't get rid of.


martinhaye:
Trying it out now.

PR #173, opened 2017-08-24
Minor changes to Series header and CMS Unit Profile

amardesi:


app/jsx/components/MediaFileGridComp.jsx
```diff
@@ -5,15 +5,10 @@ import $ from 'jquery'
...
-  url = "/content/qt" + this.props.id + "/supp/" + this.props.file
-
-  download = () => {
-    event.preventDefault()
-    window.open(this.url)
-  }
 
   render() {
     let p = this.props,
+        url = "/content/qt" + p.id + "/supp/" + p.file,
```
martinhaye: Yes! Definitely happier with this here than as an instance variable.

app/jsx/components/TabMetricsComp.jsx
```diff
@@ -3,6 +3,15 @@
 import React from 'react'
 
 class TabMetricsComp extends React.Component {
+  componentWillMount() {
+    if (!(typeof document === "undefined")) {
```
martinhaye: Excellent, thanks for keeping iso in mind!

app/jsx/components/TabMetricsComp.jsx
```diff
@@ -83,7 +92,7 @@ class TabMetricsComp extends React.Component {
             </table>
           </div>
           <div className="c-tabcontent__divide2x-child">
-            <img className="o-imagecontent" src="/images/sample_data.png" alt="sample data"/>
+            <div className='altmetric-embed' data-badge-type='donut' data-badge-details='right' data-doi="10.1038/nature.2012.9872"></div>
```
martinhaye: Shouldn't this be the item's DOI rather than a fixed DOI? Or, I guess this is temporary.

app/jsx/layouts/DepartmentLayout.jsx
```diff
@@ -1,7 +1,6 @@
 import React from 'react'
 import PropTypes from 'prop-types'
 import { Link } from 'react-router'
-import { Subscriber } from 'react-broadcast'
```
martinhaye: Thanks for keeping things clean.

app/jsx/components/TabMetricsComp.jsx
amardesi: @martinhaye Correct. And so where can I find an item's DOI?

app/jsx/components/TabMetricsComp.jsx
amardesi: @martinhaye  I wasn't sure this was the best way to bring in this JS library. It's working though.

app/server.rb
```diff
@@ -665,6 +665,28 @@ def getChildDepts(unit)
 end
 
 ###################################################################################################
+# Item Metrics 
+get "/api/item/metrics" do |shortArk|
```
amardesi: @martinhaye  This Api call hasn't been rigged up yet.  But to help with performance we could avoid grabbing this metrics .....  usage data until the user clicks on the Metrics tab. What d'ya say?

martinhaye: I agree, that makes a lot of sense. The question is how to fit in that extra API fetch. Maybe do a $.ajax call, which then fills in state?

app/jsx/components/TabMetricsComp.jsx
martinhaye: @amardesi Great question! Looks like the converter isn't converting the DOI yet... adding a story.

app/jsx/components/TabMetricsComp.jsx
martinhaye: @amardesi For this library in this context, the way you've done it seems best. It's an external library, and they want it included a certain way, and it's browser-side only, and it is only used on a tab that is seldom clicked (I would think).

PR #176, opened 2017-08-29
Info page search results using soon-to-be-defunct fulltext table. Initial integration of InfoPagesComp for preview results anyway. Got Item Metrics working.

amardesi:


app/searchApi.rb
```diff
@@ -265,6 +265,23 @@ def normalizeResponse(response)
   end
 end
 
+def getInfoResults(q)
```
amardesi: OK, I know this will all change but since I got this working with the fulltext index I figure I'll go ahead and push it for now.

PR #180, opened 2017-08-30
Big refactor of item conversion

martinhaye:


tools/convert.rb
```diff
@@ -145,6 +150,21 @@ def puts(*args)
             "3579" => "Education"}
 
 ###################################################################################################
+# Monkey-patch to nicely ellide strings.
```
amardesi: Nice invented word there Martin.

martinhaye: Heh. Merely a misspelling. :)

amardesi:
Wow. That's a lot of work. Congratulations.

PR #189, opened 2017-09-06
Informational Pages now using new index

amardesi:
Not yet complete. Waiting on Joel's work: opening up informational section with pagination.


app/listViews.rb
```diff
@@ -26,20 +27,24 @@ def itemResultData(itemIds, itemData, fields=[])
...
-            pdfCnt += 1
-          elsif supp_file['mimeType'].include?('image')
-            imageCnt += 1
-          elsif supp_file['mimeType'].include?('video')
-            videoCnt += 1
-          elsif supp_file['mimeType'].include?('audio')
-            audioCnt += 1
-          elsif ['zip', 'x-bzip', 'x-bzip2', 'x-gtar', 'x-gzip', 'gnutar', 'x-tar', 'x-zip'].include?(subtype)
-            zipCnt += 1
+          if supp_file['mimeType']
```
amardesi: I noticed an item with no mimetype in its supp_file list (First file in qt4mm9c8k3)

webpack.conf.js
```diff
@@ -1,6 +1,12 @@
 const webpack = require("webpack");
 const _ = require('lodash');
 const ManifestPlugin = require('webpack-manifest-plugin');
+const CleanWebpackPlugin = require('clean-webpack-plugin');
+
+// the path(s) that should be cleaned
+var pathsToClean = [
+  'app/js/.*-bundle-.*'
```
amardesi: This worked once and then stopped working. :(

martinhaye: I changed it to a glob (rather than a regex) and it works (though only when starting or restarting gulp): `app/js/*-bundle-*`

PR #204, opened 2017-09-22
UnitProfile bug. Anticipating campus hero image

amardesi:


amardesi:
My intention is to eventually replace each image with a campus-specific image. I'm not sure what you mean by temporary.

martinhaye:
Aren't they going to be in the CMS? In which case, they'll get asset ids and be in the S3 bucket rather than the git repo.

But I just realized the impact is currently very little -- git only stores one copy of identical binaries. So I'm just whining about nothing... :)

PR #209, opened 2017-09-27
Carousel Builder improvements; and some Joel updates

amardesi:
Carousel builder page is not yet bug-free, but figure I'd send over what amounts to the majority of the work now.


amardesi:
And thank you so much for your help!

PR #222, opened 2017-10-04
Deposit Wizard final tasks

amardesi:


app/jsx/components/StatCarouselComp.jsx
```diff
@@ -21,39 +55,45 @@ class CampusCarouselComp extends React.Component {
...
           </div>
+        }
         </div>
         <div className="c-statcarousel__section">
           <h2 className="c-statcarousel__cell-heading">
             All eScholarship
           </h2>
           <div className="o-stat--item c-statcarousel__cell">
-             <a href="">999,999</a> Items
+             <b>{a.all_item_count.toLocaleString()}</b> Items
```
amardesi: Should I wrap these in conditionals in the off chance the dbCache doesn't get populated? But if the dbCache doesn't get populated, a whole bunch of other stuff will be breaking before getting to this point anyway, yes?

martinhaye: The dbCache always gets populated before Sinatra takes the stage and binds its port. No need to wrap these in conditionals.

PR #224, opened 2017-10-05
Production performance: Puma, multiprocessing, file streaming, load test

martinhaye:


app/server.rb
```diff
@@ -95,27 +96,39 @@ def ensureConnect(dbConfig)
...
 require_relative 'citation'
 require_relative 'loginApi'
-require_relative 'fileCache'
-require_relative '../util/sanitize.rb'
-require_relative '../util/xmlutil.rb'
-require_relative 'merritt'
+require_relative 'fetch'
+
+class StdoutLogger
+  def << (str)
```
amardesi: Huh?

martinhaye: Are you asking about the syntax here? Basically this is a method that gets called when you do this:

$stdoutLogger << "stuff to log"

In my case I want to catch those occurrences and funnel them to my globally overridden "puts" which prepends the worker number and thread number to the output (and also serializes output with a mutex).

Did that answer the question?

amardesi: Yes thanks

PR #246, opened 2017-10-16
new journal cover default, ad for KF, footer tweaks.

justingonder:


martinhaye:
Merged, and pulled to dev and prd.

PR #248, opened 2017-10-16
Homepage stats; ISSN in meta tags

amardesi:


app/citation.rb
```diff
@@ -21,23 +21,24 @@ def getAuthors(a)
...
 
+  unit_attrs = unit ? JSON.parse(unit[:attrs]) : nil
   c = {
     :id => id,
     # ToDo: Put item genre here, and might also need to combine with unit type (i.e. article-journal)
-    :type => "article",
+    :type => item.genre,
     :title => item.title,
-    :URL => "http://" + request.host + "/item/" + shortArk,
+    :URL => "http://" + request.host + "/uc/item/" + shortArk,
```
amardesi: Not sure if this is the best way to do this. But rest assured: This variable is not being used anywhere as of yet, as citation has not been implemented.

PR #253, opened 2017-10-18
Vol/Iss config implementation

amardesi:
I accidentally brought this into main branch, but Justin says that's fine, to go ahead and PR it. Work is done  on this piece.  (https://www.pivotaltracker.com/story/show/151225546)

### DIFF
--- a/app/unitPages.rb
+++ b/app/unitPages.rb
@@ -156,7 +156,16 @@ end
 def getPageBreadcrumb(unit, pageName, issue=nil)
   ((!pageName and !issue) || pageName == "home") and return []
   if issue
-   return [{name: "Volume #{issue[:volume]}, Issue #{issue[:issue]}",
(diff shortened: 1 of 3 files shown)


app/unitPages.rb
```diff
@@ -156,7 +156,16 @@ def getNavBar(unit, navItems, level=1)
...
-   return [{name: "Volume #{issue[:volume]}, Issue #{issue[:issue]}",
+   vol = "Volume #{issue[:volume]}"
+   iss = "Issue #{issue[:issue]}"
+   if !issue[:numbering]
+     name = vol + ", " + iss 
+   elsif issue[:numbering] == "volume_only"
+     name = vol 
+   else
+     name = iss 
+   end
```
amardesi: Duplicative code to what is in JournalLayout.jsx. But since Breadcrumb is generated on Ruby side, this was the easiest approach.

PR #271, opened 2017-10-27
Sitemap generation (but still need index sitemap)

amardesi:
Martin, this is in a good place now for you to try it out. 

Shucks, I just realized item sitemaps are missing pointers to the PDFs themselves. It only includes pointers to item pages. 

I'm happy to manually build an index sitemap when I get in to the office Friday. I could not get that part working with this gem, so I figure it's best to write it up manually since Lisa's anxious to get it pushed to prod. Feel free to do yourself if you prefer.

Note the comment below about pathnames for files, I'm sure you can debug the issue quickly. I'll be curious to discover what I was doing wrong.

### DIFF
--- a/tools/sitemap.rb
+++ b/tools/sitemap.rb
@@ -1,0 +1,108 @@
+#!/usr/bin/env ruby
+
+# This script creates sitemaps for Google indexing
+# As of this writing, this does not geenrate the index
+
+# Use bundler to keep dependencies local
+require 'rubygems'
+require 'bundler/setup'
+
+require 'fileutils'
+require 'sequel'
+require 'yaml'
+require 'xml-sitemap'
+
+DEST_DIR = "app/sitemaps"
+FileUtils.mkdir_p(DEST_DIR)
+
+# eschol5 database
+DB = Sequel.connect(YAML.load_file("config/database.yaml"))
+
+Sequel::Model.db = DB
+class Page < Sequel::Model
+end
+
+class Unit < Sequel::Model
+end
+
+class Item < Sequel::Model
+end
+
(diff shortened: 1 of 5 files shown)


tools/sitemap.rb
```diff
@@ -0,0 +1,108 @@
...
+# Use bundler to keep dependencies local
+require 'rubygems'
+require 'bundler/setup'
+
+require 'fileutils'
+require 'sequel'
+require 'yaml'
+require 'xml-sitemap'
+
+DEST_DIR = "app/sitemaps"
```
amardesi: This wasn't actually working for me locally. I had to set the path to be /home/andy/jschol/app...etc

tools/sitemap.rb
```diff
@@ -0,0 +1,108 @@
...
+FileUtils.mkdir_p(DEST_DIR)
+
+# eschol5 database
+DB = Sequel.connect(YAML.load_file("config/database.yaml"))
```
amardesi: Ditto

tools/sitemap.rb
```diff
@@ -0,0 +1,108 @@
...
+# ToDo: Build series sitemap
+
+# ToDo: Build monograph_series sitemap
+
+# Build static pages sitemap
+makeMap('siteMapStatic.xml', true){|map|
+  Page.each { |p|
+    url = p.unit_id == 'root' ? "/" + p.slug : "/uc/#{p.unit_id}/" + p.slug
+    map.add url
+    # ToDo: I'd like to remove the <lastmod> element here but sure how to do it
```
amardesi: xml-sitemap doesn't seem to explain how to do this

PR #280, opened 2017-11-03
Search page fixes

amardesi:


app/jsx/components/SortComp.jsx
```diff
@@ -34,6 +34,7 @@ class SortComp extends React.Component {
   }
   
   render() {
+    let rows = Math.ceil(((this.state.rows > this.props.count) ? this.props.count : this.state.rows) / 10) * 10
```
martinhaye: Not sure what this is all about, but the dropdown seems to be working, so I'll not complain.

amardesi: @martinhaye Now that the rows parameters is being preserved across different queries, more logic had to be wrapped around determining the selected option in this dropdown.

Let's say you are viewing a large set of search results - say, in the hundreds - and you've set row to 50 items. Then you go to the search box and select something with fewer results - something less than 30, let's say, 29. Submitting this 2nd query preserves the rows=50 state (as of this latest code change), but of course you can't view things in groups of 50 if this latest result gives you significantly fewer results. So in this example case, the options in the dropdown should max out at 30.

PR #286, opened 2017-11-08
2 minor adjustments per CM's request

justingonder:


app/scss/_campussearch.scss
```diff
@@ -31,7 +31,7 @@ $campussearch-max-width: 29em;
   }
 
   @include bp(screen2) {
-    font-size: 2.6vw;
+    font-size: 2.0vw;
```
martinhaye: Justin, you need to feed this back to Joel as well; otherwise it'll get overridden next time we integrate.

martinhaye:
Pushed to prd.

PR #307, opened 2017-12-06
Issue caption temporary truncation

amardesi:
@martinhaye  Do you mind pushing this to prod immediately after your review? Joel is troubleshooting the issue so this is a temporary fix to prevent the screen from blowing out. I only notice the problem on Firefox.

https://www.pivotaltracker.com/story/show/152092429


martinhaye:
Pushed to prd.

PR #308, opened 2017-12-07
Generate thumbnail from cover image if it exists

amardesi:


martinhaye:
Pushed to pub-submit-prd

PR #309, opened 2017-12-14
Big Joel Integration

amardesi:


app/jsx/components/RefineComp.jsx
```diff
@@ -0,0 +1,143 @@
...
+import Breakpoints from '../../js/breakpoints.json'
+import PubYearComp from '../components/PubYearComp.jsx'
+
+class RefineComp extends React.Component {
+  state={refineActive: false, drawerOpen: false}
+
+  componentWillMount() {
+    if (matchMedia) {
+      this.mq = matchMedia("(min-width:"+Breakpoints.screen1+")")
+      this.mq.addListener(this.widthChange)
```
martinhaye: Seems like we need a corresponding removeListener (or whatever matchMedia's API is) in componentDidUnmount

app/jsx/components/LazyImageComp.jsx
```diff
@@ -0,0 +1,25 @@
+// ##### Lazy Image Component ##### //
+
+import React from 'react'
+import lozad from 'lozad'
+
+class LazyImageComp extends React.Component {
+  constructor() {
+    super();
+    this.observer = lozad('.c-lazyimage');
```
martinhaye: This should probably happen in componentDidMount

app/jsx/components/LazyImageComp.jsx
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  componentDidMount() {
+    this.observer.observe();
```
martinhaye: Should probably have a corresponding unobserve in componentWillUnmount

app/jsx/components/LazyImageComp.jsx
martinhaye: Also, this should only watch the element in question. Watching all elements of class "c-lazyimage" (which is what CSS selector `.c-lazyimage` does) is inefficient at best, and nasty at worst. Witness the similar problem we had with dotdotdot in this regard.

martinhaye: Another thing (sorry there are so many things): Lozad should only run on the client; it gets an error when we do it on the server in iso mode.  The following if statement will evaluate to try when running in the browser: `if (!(typeof document === "undefined"))`

PR #310, opened 2017-12-14
Remove eScholarship string from citation_title meta tag

amardesi:
Tested this locally and it fixes the issue


martinhaye:
Pushed to prd.

PR #312, opened 2017-12-18
Latest Joel integration

amardesi:


app/jsx/pages/SearchPage.jsx
```diff
@@ -343,12 +343,12 @@ class FacetFieldset extends React.Component {
     let out = []
     for (let facet of facets) {
       if (facet.value in checked)
-        out.push(facet)
-      else if (out.length < 5)
+        out.unshift(facet)
```
amardesi: I know this is not alphabetic. That's a whole other issue I am working on. And will touch this up once that is fixed.

app/jsx/components/ScrollToTopComp.jsx
```diff
@@ -5,7 +5,10 @@ import { withRouter } from 'react-router'
 
 class ScrollToTopComp extends React.Component {
   componentDidUpdate(prevProps) {
-    if (this.props.location.pathname !== prevProps.location.pathname) {
+    let this_search = this.props.location.search
+    let prev_search = prevProps.location.search
+    if (this.props.location.pathname !== prevProps.location.pathname ||
+      (this_search.includes("start") && !this_search.includes("start=0") && this_search !== prev_search)) {
```
martinhaye: This is now working pretty well for me. Good job.

I'll note that it's slightly annoying that it scrolls to the very top of the search page, dominated by Informational Pages, when I'm stepping through Scholarly Works results, but that'd be much harder to fix.

app/jsx/pages/SearchPage.jsx
martinhaye: This is an improvement. However, it's not quite right in this situation:
1. Start with no filters, expand Department facet, then Show More
2. Scroll down and select a child without its parent, e.g. Center for the Build Environment.
3. Close the Show More box, and the selection is not visible (because I didn't select the parent).

This will be tricky to fix. I hope you're up for it, because I dread trying to figure it out.

All that said, what you've done here is an improvement so I'm planning to merge and push this as-is for now.

martinhaye:
Pushed to prd.

PR #313, opened 2017-12-20
Initial swing at prioritizing checked facets

amardesi:
Martin, give this a try and let m,e know what you think. This is not addressing the sorting of facets, just the prioritization of checked ones.


amardesi:
Nice search to test this with:
search?q=aerosol
since it includes department facets that are more then 2 levels deep.

PR #314, opened 2017-12-20
Convert.rb: Add author_hide flag

amardesi:


martinhaye:
Pushed to jschol-prd and submit-prd.

PR #333, opened 2018-02-28
Adding ExtGATracker for ucla_epss (MH: See note)

justingonder:
MH: it looked to me like the label that comes after:
{ this.runExtGATracker(
was arbitrary, so I just used 'epssTracker'
I can't find references to this string elsewhere in the code, but if I need to make a change somewhere other than here to make this work, please let me know.


amardesi:
@justingonder What you did is perfect. This name is simply used to distinguish this tracker as separate from the others, so some arbitrary name is fine, as long as it's different from the rest. GA requires that you have a "tracker name" associated with each unique property ID, when tracking more than one property.

martinhaye:
Pushed to prd.

PR #339, opened 2018-03-14
Enabling HTML for Issue Description

justingonder:
I cribbed from a similar fix in one of MH's branches (https://github.com/eScholarship/jschol/commit/4b1d1d9bd85e20303e3aefbfa0beeba9f49073ab) so it's entirely possible I missed something... but the goal was to render HTML (if present) in the Issue Description that editors provide in OJS.

### DIFF
--- a/app/jsx/components/IssueComp.jsx
+++ b/app/jsx/components/IssueComp.jsx
@@ -2,6 +2,7 @@
 
 import React from 'react'
 import LazyImageComp from '../components/LazyImageComp.jsx'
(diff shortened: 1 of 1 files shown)


martinhaye:
Justin, do you have an issue I could test with? If not, I'll dig through the database.

justingonder:
yep: there's supposed to be some italics in the issue description (which they're using to put the cover caption in b/c it's too long for the real cover caption area) here: https://escholarship.org/uc/fb

martinhaye:
Pushed to prd, reconverted all issues that have a `description` field.

PR #354, opened 2018-04-12
Enabling series admins access to profile CMS page

justingonder:
We'd blocked this before because we didn't want to enable page & widget changes before implementing clone / fork, but in the meantime AM locked those areas down at the series level. Re-enabling access so that local admins can make changes to the 'About' blurb for series.


martinhaye:
Pushed to prd.

PR #355, opened 2018-04-16
Fix series page permissions. Keep stats pages out of search engines.

martinhaye:


app/server.rb
```diff
@@ -514,7 +514,10 @@ def sanitizeFilePath(path)
     pass
   elsif request.path_info =~ %r{/stats($|/)}
     # Don't do iso on stats reports
-    generalResponse(false)
+    resp = generalResponse(false)
+
+    # Prevent stats pages from getting into Google/Bing/etc.
+    resp.sub!('<metaTags></metaTags>', '<metaTags><meta name="robots" content="noindex"></metaTags>')
```
amardesi: Bravo @martinhaye !

PR #356, opened 2018-04-23
Search Page touch-ups

amardesi:


app/searchApi.rb
```diff
@@ -38,7 +38,7 @@ def initAllFacets()
   },
   'type_of_work' => {
     'displayName' => 'Type of Work',
-    'awsFacetParam' => {buckets: ['article', 'monograph', 'dissertation', 'multimedia']},
+    'awsFacetParam' => {buckets: TYPE_OF_WORK.keys},
```
martinhaye: Good idea.

app/searchApi.rb
```diff
@@ -298,6 +302,7 @@ def normalizeResponse(response)
 #  "info_count"=>12,        <-- different
 #  "infoResults"=> ...      <-- different
 def searchByType(params, facetTypes=$allFacets.keys, search_type)
+  params.delete("q") if isQueryEmpty(params)
```
martinhaye: Whoa, super tricky. Is this the magic that searches for everything if no query is specified. Nice.

Might be worth a comment here when you get around to it.

martinhaye:
Pushed to prd.

PR #361, opened 2018-04-30
Integration: Link style in marquee; CC badge placement; header spacing

amardesi:
Martin: Note this is a branch that doesn't include changes in PR #359 ("Search: Alpha sort dept/journal facet filters", which still needs work)


martinhaye:
Pushed to prd.

PR #365, opened 2018-05-04
Deptsortbycount; ArbitraryHtml option to wrap contents with p tag

amardesi:


app/jsx/pages/SearchPage.jsx
```diff
@@ -360,72 +366,91 @@ class FacetFieldset extends React.Component {
     }
   }
```
amardesi: Logic in code that follows has been drastically changed in order to better display checked facets.

app/jsx/pages/SearchPage.jsx
```diff
@@ -542,11 +568,10 @@ class FacetForm extends React.Component {
       let filters = this.state.query.filters && this.state.query.filters[fieldName] ? this.state.query.filters[fieldName] : {}
       let facets = fieldset.facets
       return (
-        <FacetFieldset key={fieldName} index={"facetbox" + i} data={fieldset} query={filters} handler={this.changeFacet}
+        <FacetFieldset key={fieldName} index={"facetbox" + i} data={fieldset} query={filters}
+                       handler={this.changeFacet}
                        // Have first two open by default
-                       open={[0,1].includes(i)}
-                       modal={((["departments", "journals"].includes(fieldName)) &&
```
amardesi: Logic moved inside FacetFieldset component where it's more relevant.

app/jsx/pages/SearchPage.jsx
```diff
@@ -360,72 +366,91 @@ class FacetFieldset extends React.Component {
...
     /* Gather values of checked facets */
     let checkedHash = {}
     for (let filter of this.props.query.filters)
       checkedHash[filter.value] = true
-    let checked_count = Object.keys(checkedHash).length
 
     let checked = []
-    /* Can't use Generator function here since it's not supported by IE */
+    /* Collect checked facets into variable 'checked' (recursive)
+     * Can't use Generator function here since it's not supported by IE */
```
martinhaye: Is that true even when we're using Babel? But in any case, this function is nice and clear, so it's fine.

martinhaye:
Pushed to prd.

app/jsx/pages/SearchPage.jsx
amardesi: Ah I didn't take that into consideration.

PR #371, opened 2018-05-11
Fixes related to all-items query

martinhaye:


martinhaye:
Pushed to prd.

amardesi:
Thanks Martin, looks like some thing I overlooked.

PR #375, opened 2018-06-11
Advsearch

amardesi:
• Using parslet gem to parse and transform queries including labels (title:, author:, keywords:), and any boolean (AND|OR|NOT), into cloudsearch structured query
• Added 'no stemming' analysis scheme to cloudsearch for authors


Gemfile.lock
```diff
@@ -44,21 +44,22 @@ GEM
...
+    rack (2.0.5)
+    rack-protection (2.0.2)
       rack
     sanitize (4.6.4)
       crass (~> 1.0.2)
       nokogiri (>= 1.4.4)
       nokogumbo (~> 1.4)
     sequel (4.48.0)
-    sinatra (2.0.1)
+    sinatra (2.0.2)
```
martinhaye: Thanks for upgrading this.

app/jsx/components/FilterComp.jsx
```diff
@@ -54,7 +54,7 @@ class FilterComp extends React.Component {
             ) }
           </ul>
         </details>
-        {/* <a href="" className="c-filter__tips">Search tips</a> */}
+        <a href="https://help.escholarship.org/support/solutions/articles/9000148939-using-advanced-search-beta-" className="c-filter__tips">Search tips</a>
```
martinhaye: Nicely written and clear. Question: it says "The + and - operators are not officially supported by eScholarship" ... does that mean they're supported but unofficial? Shouldn't we just not support them at all?

app/searchApi.rb
```diff
@@ -228,6 +228,11 @@ def aws_encode(params, facetTypes, search_type)
   aws_params[:start] = (search_type == "items") ?
         params.dig('start', 0) ? params['start'][0] : 0
      :  params.dig('info_start', 0) ? params['info_start'][0] : 0
+  if aws_params[:query] =~ %r{\b(title:|author:|keyword:|keywords:)}
```
martinhaye: Ah, I was wondering how the new parser gets triggered. This is nice and conservative.

app/searchParser.rb
```diff
@@ -0,0 +1,114 @@
...
+  rule(:right => simple(:right)) { right }
+
+end
+
+def q_structured(q)
+  begin
+    # View parsed tree:
+    # tree = QueryParser.new.parse_with_debug(q)
+    tree = QueryParser.new.parse(q)
+  rescue Parslet::ParseFailed => error
```
martinhaye: Maybe log something here with a puts statement?

tools/updIndex.rb
```diff
@@ -8,11 +8,14 @@
...
 # Remainder are the requirements for this program
 require 'aws-sdk'
-require 'pp'
 
 # Run from the right directory (the parent of the tools dir)
 Dir.chdir(File.dirname(File.expand_path(File.dirname(__FILE__))))
 
+# I had to run script first time to define this as an analysis scheme, 
+#  then ran script again to then call it as an author option
+analysis_schemes = [ { analysis_scheme_name: "no_stemming", analysis_options: { algorithmic_stemming: "none" }, analysis_scheme_language: "en" } ]
```
martinhaye: So what process should we use to do this on stg and prd? Comment out the author option, run it, then comment back in?

app/jsx/components/FilterComp.jsx
amardesi: We don't support them at all. I'll ask Justin to remove the word official

tools/updIndex.rb
amardesi: I'll write the code up better so the whole thing does this in two phases

PR #388, opened 2018-08-07
Form now encapsulates everything on search page (Search page selector…

amardesi:
…s, pagination now working in IE11)


martinhaye:
Pushed to prd.

PR #396, opened 2018-08-15
Hide Sort by Most Popular; Series Thumbnails; Default Search box text; Code format for Sublime Text

amardesi:


app/scss/_pubpreview.scss
```diff
@@ -23,7 +23,7 @@
 
   @include bp(screen1) {
     display: inline-block;
-    flex: 0 0 150px;
+    flex: 0 0 121px;
```
amardesi: See for an example: https://escholarship.org/uc/aah_rw

martinhaye:
Pushed to prd.

PR #405, opened 2018-09-04
Nil check check (also handle new journals that have no issues publish…

amardesi:
…ed yet)


app/unitPages.rb
```diff
@@ -463,8 +463,8 @@ def getIssueNumberingTitle(unit_id, volume, issue)
 def _getIssue(unit_id, publishedIssues, volume=nil, issue=nil, display)
   if volume.nil?  # Landing page (most recent journal) has no vol/issue entered in URL path
     i = DB.fetch(_queryIssues(publishedIssues))
+    i = i.nil? ? nil : i.first
```
martinhaye: This is fine. A cute way to accomplish the same thing: `i and i = i.first` (note 'and' instead of '&&' -- the precedence makes it work all fancy-like).

martinhaye:
Pushed to prd.

PR #408, opened 2018-09-07
Empty sections

amardesi:


tools/convert.rb
```diff
@@ -1394,6 +1394,7 @@ def updateDbItem(data)
   ItemAuthor.where(item_id: itemID).delete
   ItemContrib.where(item_id: itemID).delete
   UnitItem.where(item_id: itemID).delete
+  data['dbSection'].nil? and Item.where(id: itemID).update(section: nil)
```
martinhaye: I think nulling out the section here is redundant, because `data[:dbItem].create_or_update()` saves the whole item record.

That said, this won't hurt anybody, so I'll let it go through. You might want to try without it and if that works, drop this line.

tools/convert.rb
```diff
@@ -854,7 +854,7 @@ def parseUCIngest(itemID, inMeta, fileType)
   issue = section = nil
   volNum = inMeta.text_at("./context/volume")
   issueNum = inMeta.text_at("./context/issue")
-  if issueNum or volNum
+  if inMeta[:state] != "withdrawn" and (issueNum or volNum)
```
martinhaye: It just struck me to wonder about embargoed items. Out of scope for the story we're working on though.

martinhaye:
Pushed to submit-dev and submit-prd. (I've got stg on a branch)

tools/convert.rb
amardesi: Hmm, I did try without and section ID remained in item. I'm not sure what create_or_update does.

PR #410, opened 2018-10-11
IssueComp integration

amardesi:
No longer using CSS Grid for this component. Thumbnail was being vertically stretched in Safari, plus issues below:

I had to use a javascript solution for issues 1 and 2 to solve what otherwise appeared impossible using CSS.
1) Adjust height of the description div to deal with opening an absolutely positioned caption
2) If thumbnail isn't there, title is properly positioned


martinhaye:
Pushed to prd.

PR #411, opened 2018-10-11
Don't display month and day for items with '-01-01' pub date

amardesi:


amardesi:
@martinhaye Once you approve and push this this, I will go ahead and update the PDFs in the S3 bucket using `convert.rb --splash` command

martinhaye:
Pushed to prd.

PR #414, opened 2018-10-30
Issue caption More button 'display: none' is unnecessary, as truncati…

amardesi:
…on doesn't get triggered at mobile size; Links under c-clientmarkup element now wrap if they need to


amardesi:
This is an integration of what was already done in eschol-ui

martinhaye:
Pushed to prd.

PR #415, opened 2018-11-05
Twitterfeed

amardesi:
@martinhaye Can you push this to dev so Justin can fiddle before pushing to prod?

### DIFF
--- a/app/unitPages.rb
+++ b/app/unitPages.rb
@@ -1265,7 +1265,7 @@ put "/api/unit/:unitID/sidebar/:widgetID" do |unitID, widgetID|
     attrs = {}
     inAttrs[:title] and attrs[:title] = sanitizeHTML(inAttrs[:title])
     inAttrs[:html]  and attrs[:html]  = sanitizeHTML(inAttrs[:html])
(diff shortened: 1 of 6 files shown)


app/unitPages.rb
```diff
@@ -1263,6 +1265,7 @@ def rebuildIndirectLinks(unitID, hier = nil)
     attrs = {}
     inAttrs[:title] and attrs[:title] = sanitizeHTML(inAttrs[:title])
     inAttrs[:html]  and attrs[:html]  = sanitizeHTML(inAttrs[:html])
+    inAttrs[:twitter_handle]  and attrs[:twitter_handle]  = sanitizeHTML(inAttrs[:twitter_handle])
```
martinhaye:
```suggestion
    inAttrs[:twitter_handle]  and attrs[:twitter_handle]  = inAttrs[:twitter_handle]
```
We don't render the twitter handle as HTML do we? It gets rendered as plain text, so no need to sanitize it. That said, it's probably harmless to do so.

martinhaye:
Pushed to prd.

PR #416, opened 2018-11-14
Links under clientmarkup don't need 'u-one-line-truncation' mixin

amardesi:
Fixes this problem: https://www.pivotaltracker.com/story/show/161941746


martinhaye:
Pushed to prd.

PR #422, opened 2019-01-10
Surround name with quotes to make author specific query

amardesi:


amardesi:
Examples to test with:
https://www.pivotaltracker.com/story/show/158773784
https://www.pivotaltracker.com/story/show/159857693

app/jsx/components/DrawerComp.jsx
```diff
@@ -40,24 +40,38 @@ class SortableNavList extends React.Component {
...
+        return undefined
+      return navItems.map(nav => {
+        let data = {
+          id: nav.id,
+          type: nav.type,
+          slug: nav.slug,
+          title: <Link to={`/uc/${this.props.unit}${(nav.type.includes("fixed")) ? "" : `/nav/${nav.id}`}`}>{nav.name}</Link>,
+          subtitle: <i>{nav.hidden && 'hidden '}{nav.type.replace(/_/g," ")}</i>
+        }
+        if (nav.type.includes("folder") && (nav.id >= 0 || nav.id == -9999)) {
```
martinhaye: This could use a comment. I think I know what the >= 0 || == -9999 is about (avoiding the issues sub-nav) but I had to really think about it.

app/jsx/components/DrawerComp.jsx
```diff
@@ -78,23 +92,23 @@ class SortableNavList extends React.Component {
...
           return true
         }}
         canDrop={({ node, nextTreeIndex, nextParent }) => {
-          if (this.state.firstIsHome && nextTreeIndex == 0) // don't allow rearranging above the unit home
+          if (this.state.firstIsFixed && nextTreeIndex <= 0) // don't allow rearranging above fixed nodes
             return false
           if (!nextParent)
             return true
-          if (node.type == "folder")
+          if (node.type.includes("folder"))
```
martinhaye: Why the change from == "folder" to includes("folder")? Are there nodes of some larger string which contains "folder" that need to be treated as folders? Maybe this question will be answered as I continue reading...

app/unitPages.rb
```diff
@@ -637,7 +702,7 @@ def getUnitSidebarWidget(unit, widgetID)
 def travNav(navBar, &block)
   navBar.each { |nav|
     block.yield(nav)
-    if nav['type'] == 'folder'
+    if nav['type'] and nav['type'].include?('folder')
```
martinhaye: Works fine, but in general "and" in this context should be "&&"; I think we conventionally save "and" for an expression on the left that conditionally triggers a command on the right. But please correct me if I'm off-base here.

app/unitPages.rb
```diff
@@ -653,6 +718,7 @@ def getNavByID(navBar, navID)
 def deleteNavByID(navBar, navID)
   return navBar.map { |nav|
     nav['id'].to_s == navID.to_s ? nil
+      # Check for 'folder'    (You would never want to delete a 'fixed_folder')
```
martinhaye: Ah, there's the answer to my question. "fixed_folder" is the other kind of folder. Ok.

PR #423, opened 2019-01-11
Adding directSubmitURL field to Unit Profile CMS page

justingonder:
Never tried adding something like this, so check carefully and throw away if I royally screwed up ;)
per https://www.pivotaltracker.com/story/show/163153782


martinhaye:
Changes made, pushed to prd.

PR #440, opened 2019-03-22
Force any queries on author to be phrase queries; Lastname like 'And'…

amardesi:
… or 'Or' confused as logic statements


martinhaye:
Pushed to prd.

PR #480, opened 2020-03-20
Removing entire 'Citation' level from Download Menu

justingonder:
Other as of yet unsupported reference links in the download menu were already commented out, but EndNote wasn't and needed to be. Removing entire section until we're ready to implement these.


martinhaye:
Tested, and it works nicely. Will put into production shortly.

martinhaye:
Deployed to prd.

PR #483, opened 2020-04-24
PT-172061580 Add a fix for garbled UTF-8 characters to convert.rb

hardyoyo:
* Add a workaround for [a bug in mutool](https://bugs.ghostscript.com/show_bug.cgi?id=702358)
which produces two-digit hex escape strings instead of UTF-8 characters
* This workaround uses a gsub method (which is a destructive regex) and
forces encoding to UTF-8
* This gsub line is the work of @martinhaye he deserves all the credit
in writing it.

### DIFF
--- a/tools/convert.rb
+++ b/tools/convert.rb
@@ -698,6 +698,8 @@ def generatePdfTOC(itemID)
     divs = []
     buf = ""
     outline.each { |line|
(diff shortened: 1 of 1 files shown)


hardyoyo:
This code needs testing before I make a PR, so I closed this until that test is complete.

PR #493, opened 2020-10-06
Update TabMetricsComp.jsx

justingonder:
Per user request: Clarifying language that gets displayed when no AltMetric or Dimensions information is available for display on an item's Metrics tab.


hardyoyo (approved):
👍  obviously fine

PR #496, opened 2020-12-04
hide campus selector and breadcrumb for Combinatorial Theory as per Pivotal story 176017264

hardyoyo:


martinhaye (commented):
Looks fantastic!!!!!

PR #501, opened 2021-03-30
dermatology / repository hostnames linger on some pages that should redirct 

tingletech:
still needs to be tested

### DIFF
--- a/app/redirect.rb
+++ b/app/redirect.rb
@@ -44,6 +44,10 @@ def checkRedirect(origURI)
            "pub-jschol-prd.escholarship.org"
           ].include?(uri.host)
       uri.host = "escholarship.org"
+    elsif uri.host == "repositories.cdlib.org"
+      uri = handleBepressRedirect(uri)
+    elsif uri.host =~ /dermatology(-s10)?.cdlib.org/
+      uri = handleDojRedirect(uri)
     elsif uri.path =~ %r{^/uc/oai/?(.*)}
       uri.path = "/oai#{$1}"
       break
@@ -61,10 +65,6 @@ def checkRedirect(origURI)
       break
     elsif uri.path =~ %r{^/uc/([^/]+)(.*)}
       uri = handleUnitRedirect(uri, $1, $2)
-    elsif uri.host == "repositories.cdlib.org"
-      uri = handleBepressRedirect(uri)
-    elsif uri.host =~ /dermatology(-s10)?.cdlib.org/
-      uri = handleDojRedirect(uri)
     elsif uri.path =~ %r{^/oa_harvester/}
       uri.path = "/images#{uri.path}"
       break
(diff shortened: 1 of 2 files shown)


hardyoyo:
The modified test fails in my local dev environment:
```
lando ruby test/testRedirects.rb
...........................Error: Incorrect redirect. "http://dermatology.cdlib.org/uc/doj/14/3" -> "https://escholarship.org/uc/doj" but expected "http://escholarship.org/uc/doj/14/3"
```

app/redirect.rb
```diff
@@ -44,6 +44,10 @@ def checkRedirect(origURI)
            "pub-jschol-prd.escholarship.org"
           ].include?(uri.host)
       uri.host = "escholarship.org"
+    elsif uri.host == "repositories.cdlib.org"
```
hardyoyo, line 47: In a conversation with @martinhaye earlier today, I think he was suggesting that the order here is important, and we may want to find another place for these so that your new tests pass.

hardyoyo (changes requested):

app/redirect.rb
tingletech: obviously the order is important.  but, judging from the failure, I think the problem in now in `handleBepressRedirect` (which probably needs a test too) and `handleDojRedirect`, is it not, because as far as I see that is the only place that would send a redirect to https://escholarship.org/uc/doj

With the previous order, the handle...Redriect's were not even getting called for this URL

hardyoyo: I've given this a couple hours today, trying to rewrite the logic in the checkRedirect method so that the new test will pass, in order to move this along. I'm afraid I can't really trace the existing logic at all, so, I'm throwing in the towel.  In the role of reviewer of a PR, I can only say: this suggested change does not pass the provided test.

tingletech:
Right now, this is there the test is failing

```
Error: Incorrect redirect. "http://dermatology.cdlib.org/uc/doj/14/3" -> "https://escholarship.org/uc/doj" but expected "http://escholarship.org/uc/doj/14/3"```

I think that the best thing to do at this point is to change the test to expect `https://escholarship.org/uc/doj`

We also need to empirically test where `http://repositories.cdlib.org/uc/doj/14/3` goes on an instance running this code, and then put that in the test script.

hardyoyo:
> Right now, this is there the test is failing
> 
> ```
> Error: Incorrect redirect. "http://dermatology.cdlib.org/uc/doj/14/3" -> "https://escholarship.org/uc/doj" but expected "http://escholarship.org/uc/doj/14/3"
> ```
> 
> I think that the best thing to do at this point is to change the test to expect `https://escholarship.org/uc/doj`

That sounds reasonable to me.

> We also need to empirically test where `http://repositories.cdlib.org/uc/doj/14/3` goes on an instance running this code, and then put that in the test script.

I can test that now, give me a few minutes.

hardyoyo:
confirmed tests now pass, as far as I'm concerned this is ready to merge

hardyoyo:
@justingonder you and @tingletech should confer on whether this PR is the "right" solution, from a code review standpoint, it's ready to go

hardyoyo (approved):

hardyoyo:
Closing this pull request, we're going to approach the duplicate URL indexing in Google issue in a slightly different way.

PR #502, opened 2021-03-30
Adding text alignment button to TrumboWYSIWYG

justingonder:


hardyoyo:
testing this PR in my local dev environment, I cannot see an alignment button in the WYSIWYG editor

hardyoyo:
In reviewing the Trumbow docs, I discovered there is no such button `align` however you *can* create a custom definition for a drop-down button and call it 'align'. There are some default buttons for alignment:

https://alex-d.github.io/Trumbowyg/documentation/#button-pane

And I can confirm that they work fine.

hardyoyo:
replaced by #506

PR #505, opened 2021-04-21
Remove old/unused Docker stuff

hardyoyo:
Removes the old docker stuff, which has since been replaced with Lando.


hardyoyo:
We can always revert to get them back if we want them, and we don't use them, so I'm merging this.

tingletech:
Makes sense

PR #506, opened 2021-04-22
PUBD-17 Enable JustifyRight (and other text alignment) in Trumbo

hardyoyo:
* Add the default justification buttons to the TRUMBO_BUTTONS constant for TRUMBO
* Allow the style attribute for p tags in sanitizeHTML
* Supercedes #502 
* contains some benign dependency version bumps for these gems: nokogiri, nokogumbo and sanitize


tingletech (approved):
👍

PR #510, opened 2021-05-04
PUBD-5 and PUBD-148: only attempt to normalize DOIs that exist

hardyoyo:
- Only attempt to normalize a DOI if a DOI exists, otherwise skip it (the regex match doesn't like nulls)


martinhaye (approved):
Oh, good point!

tingletech (approved):

PR #512, opened 2021-06-15
switching to erubis implimentation of eRuby

tingletech:


hardyoyo:
More info on Erubis: http://www.kuwata-lab.com/erubis/ This patch mostly uses the auto escaping feature, but, we might explore some of the other features that Erubis brings, in the future.

hardyoyo (commented):
👍 looks good, and passes tests in my local environment.

PR #514, opened 2021-06-16
Escape search param values as they are parsed out of the query string

hardyoyo:
- Escapes the search terms as they are parsed out of the query string
- Carefully resets the default of the cgiParams hash to [] instead of
nil, to keep the "magic" in-tact


hardyoyo:
This has been running on my local dev instance, searches work as before (the quicktest proves this), and searches with special characters are escaped as they should be.

tingletech (approved):

PR #515, opened 2021-06-22
PUBD-201 fix advanced search

hardyoyo:
Unescapes search parameters where necessary (on search param display of search results, and before search params are sent to CloudSearch). Restores adavanced search functionality where it was broken due to the presence of escaped search params in the CloudSearch request.


martinhaye:
Nice job tracing the culprit here.

tingletech (commented):
A unit test for `author:"Atkinson, Richard"` might be good

tingletech (approved):

hardyoyo:
> Nice job tracing the culprit here.

Credit where credit is due, @tingletech did all the real sleuthing here. I just smashed all the bits with a hammer until it worked.

PR #525, opened 2021-08-03
2021.08 admin template updates

alainna:
Update admin template: 
* Monthly -> Regular 
* Strength language on how to increase traffic, why receiving, how to notify of role change


myucekul (approved):
The changes look good.

PR #526, opened 2021-08-03
Updated robots pattern list

myucekul:


hardyoyo:
Note: the Rubocop lint check will always fail on PRs from public forks (such as this one). Not a big deal.

hardyoyo (approved):
looks fine to me

PR #527, opened 2021-08-03
Comment out the Rubocop lint check, it doesn't work for public forks

hardyoyo:
Since the Rubocop check always fails on pull requests from public forks, and we're planning to replace Github actions with AWS Codebuild for our CI, we should disable the Rubocop lint check for now.


hardyoyo:
Discussed this during a recent PAD-Tech meeting, and no objections were raised. Merging this PR

PR #528, opened 2021-08-05
Add link rel=canonical to MetaTagsComp invocations on GlobalStaticPage, StatsPage and UnitPage

hardyoyo:
- Uses location.pathname from props to set the canonical link, and hard codes the domain to escholarship.org with https as the transport


hardyoyo:
Passes tests locally.

tingletech (approved):
seems like that would work

PR #529, opened 2021-08-11
removed - require 'mimemagic/overlay'

myucekul:
removed - require 'mimemagic/overlay'


hardyoyo:
Passes tests locally.

PR #532, opened 2021-09-20
Run lint checks before the quick test via Lando test tooling

hardyoyo:
A minor revision to the test tooling for Lando, to ensure lint checks are run before the code is tested.


martinhaye (approved):
Good idea!

PR #535, opened 2021-09-27
PUBD-254: remove tabIndex from all jsx pages

hardyoyo:


martinhaye (commented):
I see tabIndex="-1" on lots of components *other* than pages. Joel only wants it removed on `<main>`?

--Martin

hardyoyo:
I assume so, based on this commit: https://github.com/eScholarship/eschol-ui/commit/237c641cd3c090a2bf645fe2923c811fc20a24cf

> I see tabIndex="-1" on lots of components _other_ than pages. Joel only wants it removed on `<main>`?
> 
> --Martin

martinhaye:
Yep, you're right.

PR #538, opened 2021-10-06
Fixes PUBD-299, a pagination error in Jschol

hardyoyo:
* Ensures the hidden form field in FilterComp that holds the search query is properly unescaped (and usable)
* Adds a test that confirms that an author search results in a properly escaped hidden form field in FilterComp


hardyoyo:
passes tests locally

PR #545, opened 2021-11-30
PUBD-336: actually link to the PDF, and add popup for restricted items

hardyoyo:
Change the link to the PDF, and use the same popup logic for restricted items as is present for the download button.


hardyoyo:
I've confirmed with @alainna that the existing logic for the restricted item popup is known to no work, so this PR is "as good as what's on prd". I've marked this as ready for review.

myucekul (commented):
for the download restricted case, is it better to point href to the download link or to keep it empty?

myucekul (approved):

PR #546, opened 2021-12-02
PUBD-390 Port heading 5 styles from eschol_ui

hardyoyo:
As per PUBD-390, I copied the changes from https://github.com/eScholarship/eschol-ui/commit/6e3843f98701e2f7d2c4a1e1e9da10e4e1991384


myucekul (approved):
thanks for the change!

PR #548, opened 2021-12-03
Improve quickest, allowing it to target other servers

hardyoyo:


martinhaye (approved):
Seems like a handy improvement

PR #550, opened 2021-12-06
PUBD-370 Add semicolons to AuthorsComp in RecentArticlesComp

hardyoyo:
Conditionally adds semicolons to the AuthorsComp in RecentArticlesComp (not to be confused with the similarly-named AuthorListComp)


myucekul (approved):

myucekul (commented):
So, the separator is changing from , to ;. what about the separator before et al?
 out += _.map(auths.slice(0,6), this.formatAuth).join("; ") + ", et al."

hardyoyo:
> So, the separator is changing from , to ;. what about the separator before et al? out += _.map(auths.slice(0,6), this.formatAuth).join("; ") + ", et al."

That code is for building citations, and isn't responsible for the author list on item or unit pages. The "et al. problem" will have to remain for another day.

PR #551, opened 2021-12-08
changes for running jschol on AL2

myucekul:


deployVersion.sh
```diff
@@ -61,13 +61,13 @@ npm install
 
 # Pretranslate all the CSS
 echo "Building app."
-./gulp sass
+./node_modules/.bin/gulp sass
```
hardyoyo, line 64: Lando is currently set up to use the former symlink, but it can be changed to use the full path to gulp

myucekul: Yes, please point to the file in node_modules. This change is needed because of the order in which hooks are executed and unzip is performed. I moved these links to nodein.sh

hardyoyo: I'm not currently able to get a Lando development environment working, with these changes added. I will keep trying tomorrow, to see if I can make any suggestions.

.lando.yml
```diff
@@ -29,19 +30,21 @@ services:
...
       - apt-get install -y vim net-tools
-      - curl -sL https://deb.nodesource.com/setup_10.x | bash -
+      - curl -sL https://deb.nodesource.com/setup_10.x | sed '/sleep/d' | bash -
       - curl -sS https://dl.yarnpkg.com/debian/pubkey.gpg | apt-key add -
       - echo "deb https://dl.yarnpkg.com/debian/ stable main" | tee /etc/apt/sources.list.d/yarn.list
       - apt-get update -qq
       - apt-get autoclean
       - apt-get install mariadb-client build-essential libpq-dev yarn nodejs -y
       - gem install bundler
+      # - gem install foreman
```
hardyoyo, line 40: sorry, I missed this commented out line, it can go

deployVersion.sh
```diff
@@ -32,7 +32,7 @@ VERSION+=`git rev-parse --short HEAD`
 DIR=jschol
 BUCKET=cdlpub-apps
 REGION=us-west-2
-APPNAME=eb-pub-jschol
+APPNAME=eb-pub-jschol2
```
hardyoyo, line 35: will we retain this name-change going forward?

tools/convert.rb
```diff
@@ -18,7 +18,7 @@
 require 'aws-sdk-s3'
 require 'date'
 require 'digest'
-require 'ezid-client'
+#require 'ezid-client'
```
hardyoyo, line 21: should this just be deleted?

hardyoyo (approved):
Two small comments/suggestions, but I have no objections to merging this.

hardyoyo (approved):
passes tests on the jschol2 beanstalks, 👍

PR #552, opened 2021-12-09
PUBD-416: Tame MathJax rendering issues

hardyoyo:
- Add call to MathJax.typeset() in a componentDidMount, which should force
the rendering of MathJax, if there's any to render

### DIFF
--- a/app/js/mathjax-typeset.js
+++ b/app/js/mathjax-typeset.js
@@ -1,0 +1,1 @@
+MathJax.typeset();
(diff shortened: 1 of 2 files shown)


myucekul (commented):
where is the function componentDidMount used?

hardyoyo:
> where is the function componentDidMount used?

docs: https://www.educba.com/react-componentdidmount/
inspiration to use: https://stackoverflow.com/questions/40958680/react-router-how-to-trigger-document-ready (hat tip to @tingletech for finding that)

PR #559, opened 2022-01-11
PUBD-416 Make MathJax Local, and SSR friendly

hardyoyo:
- Improve MathJax performance by moving the library to a self-hosted version
- Improve stability of MathJax integration by preventing it from being loaded by the Isometric/SSR process
- Remove unused MathJax calls, as well as an unrelated yet likewise unused import
- Tune up the regex used to detect content that requires MathJax, to make the regex performant enough for prd


myucekul (commented):
Hardy - do you want to swap the conditions in app/jsx/components/ArbitraryHTMLComp.jsx?
That should mean ISO doesn't need to look for dollar..
if (!(typeof document === "undefined") && /\$[^\d]/.test(fixedText2))

PR #561, opened 2022-01-13
Refactor out unneccessary and potentially slow regex matches

hardyoyo:
- Replace two unneccessary regex matches with startsWith and endsWith


tingletech (approved):
makes sense to me

PR #566, opened 2022-02-04
Sortseries 59

myucekul:


everreau (commented):
Looks OK to me.

hardyoyo (approved):
This looks correct, passes tests locally, and when I compare the sorting on this page:
https://escholarship.org/uc/sio
To the one running on my local dev environment, the sort order is indeed different.

So 👍 from me.

PR #568, opened 2022-02-09
Added sort by ordering for siblings on dept landing page and sort by …

myucekul:
…name for children in units page


hardyoyo (approved):
I cannot parse the contents of the PUBD-59 card, so I cannot verify whether this work satisfies the request. However, I *can* say that, in my local dev environment, the tests pass... so, that has to count for something. :+1: because I trust @myucekul, and this doesn't appear to be broken

PR #572, opened 2022-03-10
Rename assets to cms-assets to avoid EB webapp conflict with the url

myucekul:


hardyoyo (approved):
👍  looks good, and has been tested on the JSchol2 dev beanstalk

PR #574, opened 2022-03-15
PUBD-505 Add a test for broken images

hardyoyo:
Adds a test for broken images, closes PUBD-505

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,7 +1,7 @@
 ---
 
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 3.0
   Exclude:
     - 'gems/**/*'
     - 'node_modules/**/*'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -29,6 +29,8 @@ GEM
     builder (3.2.4)
     crass (1.0.6)
     daemons (1.4.1)
+    domain_name (0.5.20190701)
+      unf (>= 0.0.5, < 1.0.0)
     em-websocket (0.3.8)
       addressable (>= 2.1.1)
       eventmachine (>= 0.12.9)
@@ -39,6 +41,9 @@ GEM
     get_process_mem (0.2.7)
       ffi (~> 1.0)
     htmlentities (4.3.4)
+    http-accept (1.7.0)
+    http-cookie (1.0.4)
+      domain_name (~> 0.5)
     httparty (0.20.0)
       mime-types (~> 3.0)
       multi_xml (>= 0.5.2)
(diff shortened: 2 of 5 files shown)


myucekul (commented):
I'm concerned about Gemfile changes for this test. Can test be independent of Gemfile used for EB?

myucekul (commented):
Can we also test for other assets links - like on submission page from slack messages about doj submissiongruidelines today.

.rubocop.yml
```diff
@@ -1,7 +1,7 @@
 ---
 
 AllCops:
-  TargetRubyVersion: 2.5
+  TargetRubyVersion: 3.0
```
tingletech, line 4: why would this change for a test?

Gemfile.lock
```diff
@@ -169,4 +184,4 @@ RUBY VERSION
    ruby 2.6.8p205
 
 BUNDLED WITH
-   2.2.27
+   2.3.8
```
tingletech, line 187: why would this change for a test?

.rubocop.yml
hardyoyo: That's the version of Ruby we run on the beanstalks, and we run Rubocop along with our tests.

Gemfile.lock
hardyoyo: I'll drop this change, it's not necessary.

hardyoyo: actually, it might be necessary, I had to add a few gems to facilitate the new test

hardyoyo: I added at least one new gem to facilitate the image test, and ran bundle to install the gems. This is the result.

hardyoyo:
> I'm concerned about Gemfile changes for this test. Can test be independent of Gemfile used for EB?

I had to add a couple gems to facilitate the new test for broken images. They're only used in the test, and should not impact the beanstalk at all.

hardyoyo:
> Can we also test for other assets links - like on submission page from slack messages about doj submissiongruidelines today.

Yes, though the card PUB-505 does not currently mention such a requirement. My inclination would be to accept this test first, then improve it as part of a new card.

hardyoyo:
> > Can we also test for other assets links - like on submission page from slack messages about doj submissiongruidelines today.
> 
> Yes, though the card PUBD-505 does not currently mention such a requirement. My inclination would be to accept this test first, then improve it as part of a new card.

Here's where we pull all the images out of the page we're testing:
https://github.com/eScholarship/jschol/pull/574/files#diff-8d97493c1f164c1a62f29f18a4e5b5ad7b4d6424854f8629d1d14f9c3296aaaaR174

We'd need to be able to search for the links inside the document, so we'd need to figure out the correct criteria for the search. That would be part of the future card we need to write up. We should probably also refactor this test to make the bulk of it into a function, so we can test multiple URLs for multiple kinds of things (images, other links, etc.).

tingletech (approved):

hardyoyo:
Card [522](https://trello.com/c/r5xkFJtU) has been created to track the desired improvements mentioned above.

PR #581, opened 2022-05-12
Upgrade Node to v14

hardyoyo:
- Upgrades the version of Node installed for Lando, as well as for beanstalks, to v14
- Updates dependencies to work with Node 14
- Removes deprecated babel-polyfill, at the suggestion of npm update
- adds the aws folder to eslint ignore file, not our business
- Adjusts broken image test to check a unit page that exists on dev
- Incorporates the Gemfile changes from dependabot pull requests


hardyoyo:
This branch builds and passes tests locally. Needs to be tested in Elastic Beanstalk. But is ready for review.

hardyoyo:
this code has been deployed to dev for further testing

hardyoyo:
For testing, @martinhaye has suggested that we compare results for a few different browsers and platforms, since this is a comprehensive update of all or most of the dependencies (with one deprecated dependency completely removed).

hardyoyo:
Results of testing dev via crossbrowsertesting.com: the site works on all browsers tested, and looks mostly correct. IE looks weird, the buttons are *huge* I saved a screenshot. I doubt that's specific to this PR, though. And the test of the iPad Air version was unusably sluggish, which I attribute to a problem with the crossbrowsertesting site, but it would be worth a re-test tomorrow.

myucekul (approved):

PR #586, opened 2022-05-25
Fix promoteVersion script with changes from deployVersion

hardyoyo:
Stability changes
- replace shell preset variables with named variables
- use improved evironment existence check from deployVersion


hardyoyo:
This PR is stale, and the changes made it into master from another branch. Closing.

PR #592, opened 2022-06-30
Render all math content on page load, simplify configuration

hardyoyo:
- re-enbable typesetting of math content on page load, which is default MathJax
- abandon use of regex the conditionally render MathJax, no longer needed
- use new delimiters for MathJax


hardyoyo:
this is stale, closing

PR #593, opened 2022-07-15
PUBD-396 Add editors and advisors to lists and search results

hardyoyo:
Add editors and advisors to lists and search results. Re-use author markup for editors and advisors. May need further design refinement.


myucekul (commented):
The changes look good. I noticed that the labels are all singular (author, editor etc.) but the expectation is that each one can be a list. I assume the labels are as per Program Managers ask.

hardyoyo:
> The changes look good. I noticed that the labels are all singular (author, editor etc.) but the expectation is that each one can be a list. I assume the labels are as per Program Managers ask.

Screenshot_20220715-204835-885

app/jsx/components/AuthorListComp.jsx
```diff
@@ -17,42 +16,60 @@ class AuthorListComp extends React.Component {
...
+      } 
     })
   }
 
   render() {
     let p = this.props,
-        year = p.pubdate.match(/\d{4}/),
-        source = p.source,
+        year = p.pubdate ? (p.pubdate.match(/\d{4}/)) : null,
+        source = p.source ? p.source : null,
```
myucekul, line 34: Isn't 'source = p.source ? p.source : null' same as 'source = p.source'

myucekul (approved):

app/jsx/components/AuthorListComp.jsx
hardyoyo: you're right, I was trying to rule out some possibilities, but I'll fix this

PR #595, opened 2022-07-27
Updated logic to allow for non-dollar MathJax delimiter

myucekul:


hardyoyo (approved):
As I mentioned in a conversation earlier, this regex is likely moot anyway, based on my current understanding of how Better React Mathjax works. We're always getting MathJax on the entire page, due to the MathjaxContext on PageBase.jsx. I'm glad that removing this regex gets things working (I've seen it, I believe), and have no reason to say no to this PR, so, I approve!

PR #599, opened 2022-08-02
Invoke Mathjax.typeset on tab changes

hardyoyo:


hardyoyo:
This change may not be necessary, see this comment https://github.com/fast-reflexes/better-react-mathjax/issues/22#issuecomment-1205585799

We may have inadvertently caused the issue this PR attempts to address... by removing the MathJax component. The observed behavior closely resembles what is described in that comment.

hardyoyo:
this is stale, closing

PR #603, opened 2022-08-22
Adding script for 2022 data sync from prod

myucekul:


hardyoyo (approved):
This looks fine to me... I worry about putting real e-mail addresses in public GitHub repositories, but this is your call. I also worry about leaving in commented-out lines, but I understand they can be useful suggestions for future debugging. Neither of these worries translate to suggested changes, so, I approve as-is.

PR #604, opened 2022-09-06
PUBD-674 remove et al and use dot dot dot for truncating contributor names on non-item pages

hardyoyo:
- Remove 'et al' line altogether
- rely on dot dot dot to truncate authors/contributor names

Currently deployed to dev for review purposes.

### DIFF
--- a/app/jsx/components/AuthorListComp.jsx
+++ b/app/jsx/components/AuthorListComp.jsx
@@ -61,12 +61,10 @@ class AuthorListComp extends React.Component {
         {/* CASE: this is NOT an item page */}
         { ((authors && authors.length > 0) || (editors && editors.length > 0) ||
            (advisors && advisors.length > 0)) && ! this_is_a_tabbed_page &&
-          <TruncationObj element="ul" className="c-authorlist__list"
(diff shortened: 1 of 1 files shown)


app/jsx/components/AuthorListComp.jsx
```diff
@@ -52,20 +53,18 @@ class AuthorListComp extends React.Component {
           {editors}
           {advisors}
           {/* Note: the <a> more-link below cannot be a <Link>, else jquery dotdotdot can't recognize it */}
-
-          { <li><a href={item_link} className="c-authorlist__list-more-link">et al.</a></li> } 
+          {/* Kludge: only show et al if the total number of contributors is greater than 6 */}
+          { (total_contributors > 6) &&  <li><a href={item_link} className="c-authorlist__list-more-link">et al.</a></li> }
```
hardyoyo, line 57: I did want to call out this one remaining instance of 'et al.' which matches the same logic as what's used in generating the standard citation (i.e. only use 'et al.' when there are more than 6 contributors). This seems safe, and using it only on the item page makes sense.

myucekul (approved):
Looks good

PR #606, opened 2022-09-13
PUBD-736 Add and use ESCHOL_API_ACCESS_KEY if necessary to access dev or stg API

hardyoyo:
Adds and uses a new ENV, ESCHOL_API_ACCESS_KEY. If this ENV is present, it is added to the API request in the submitAPIMutation method. Currently, the only operation in Jschol that uses the submitAPIMutation method is the withdrawal request, which is what PUBD-736 concerns. However, if other uses of the submitAPIMutation are added in the future, they will also benefit from this addition.  ESCHOL_API_ACCESS_KEY is not required on prd, only dev and stg need it.

Big thanks to @everreau for working with me on this solution! We tried **many** different ways to get the access key in to the API request, and finally found the right way to do it.


everreau (approved):
looks good!

myucekul (approved):
Looks good.

PR #607, opened 2022-09-21
PUBD-744 Upgrade to Node 16

hardyoyo:
- Upgrade to use Node 16 in our beanstalk, our Lando dev environment and our CodeBuild config.
- Use Ruby 3.0 in Codebuild via rbenv (revert from the Ruby 3.1 version provided)


buildspec.yml
```diff
@@ -5,15 +5,17 @@ version: 0.2
 phases:
   install:
     runtime-versions:
-      nodejs: 14
-      ruby: 2.7
+      nodejs: 16
+      ruby: 3,1
```
myucekul, line 9: I wonder why this says 3,1? You mean 3.0?

gulpfile.js
```diff
@@ -5,7 +5,7 @@ const del = require('del')
 const fs = require('fs')
 const gulp = require('gulp')
 const gutil = require('gulp-util')
-const sass = require('gulp-sass')
+const sass = require('gulp-sass')(require('sass'))
```
myucekul, line 8: Would you please elaborate on why sass was not required before and now is?

package.json
```diff
@@ -89,6 +90,7 @@
     "jquery.dotdotdot": "github:amywieliczka/jQuery.dotdotdot#4f7d03a442ceca0b75a2fb641a065e47297d5666",
     "lodash": "^4.17.21",
     "lozad": "^1.16.0",
+    "npm": "^8.19.2",
```
myucekul, line 93: npm is added because of code build requirement?

buildspec.yml
hardyoyo: that comma was a typo (fixed now), however, the runtime from CodeBuild only includes Ruby 3.1, so that's what we have to ask for.

gulpfile.js
hardyoyo: I will provide a link when I can find it in my browser history, however, this is the official way you have to install gulp-sass

package.json
hardyoyo: This is probably something that came along with Node 16, I honestly don't know why npm version *wasn't* tracked in Package.json before.

gulpfile.js
hardyoyo: https://stackoverflow.com/questions/68417640/gulp-sass-5-does-not-have-a-default-sass-compiler-please-set-one-yourself

myucekul (approved):

myucekul (approved):

hardyoyo:
This PR, after it is merged, will create the necessity for PUBD-758 (update Node on pub-cattle2-ops)

PR #609, opened 2022-09-21
Stats query update to account for mysql failures

myucekul:
Related to PUBD-737


hardyoyo (approved):
Many places in Jschol have comments that summarize the logic that follows; this might be a nice place to add such a comment, to explain the date math in this query. I'll approve this anyway, but thought I'd mention it. I'll wait to merge a bit, in case you want to add a comment.

hardyoyo:
I'm going to merge this as-is, we talked about it on Slack, you'll add a comment later.

PR #616, opened 2022-11-16
New CMS fields for journals - Pubd690

myucekul:


hardyoyo (approved):
This all looks good to me. I'll deploy it to dev so we can give it a proper test.

PR #622, opened 2022-12-07
Refresh cache every hour

myucekul:
Removed the logic that doesn't work and replaced it with every hour cache update.

### DIFF
--- a/app/dbCache.rb
+++ b/app/dbCache.rb
@@ -219,10 +219,10 @@ end
 Thread.new {
   begin
     while true
-      sleep 60*3600  # re-fill every 60 min
+      sleep 3600  # re-fill every 60 min
       # if the cache was filled otherwise - skip refilling it from this thread
-      if ((Time.now - $lastFillTime)/3600 > 30)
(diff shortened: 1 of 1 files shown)


everreau:
Looks OK to me.

app/dbCache.rb
```diff
@@ -212,32 +213,15 @@ def signalDbRefill
...
-  return utime
-end
-
 ##################################################################################################
 # Database caches for speed. We check every 30 seconds for changes. These tables change infrequently.
 
 Thread.new {
   begin
-    prevTime = getLastTableUpdateTime
+    puts "IN DB CACHE Thread"
```
hardyoyo, line 221: leftover debug logging?

hardyoyo: I'd suggest changing this text to something slightly more meaningful if we want to keep it, such as "ENGAGED: 60 minute DB cache refill timer" or similar.

hardyoyo (changes requested):

hardyoyo (approved):

PR #596, opened 2022-07-28
Get rid of SOCKS-mysql kludge

martinhaye:
The kludge broke compatibility with MySQL 8.0 and appears to not be needed with 5.7 either. Tested with dev (8.0) and stg (5.7).

PR #625, opened 2022-12-14
Revert #596, add back the kludge with updated byte values

hardyoyo:
- turns out we need the kludge in order to sucessfully connect to MySQL
  via a SOCKS proxy
- this is necessary for some dev environments (specifically ones that
  use a SOCKS proxy to connect to a database)


myucekul (commented):
looks okay.

PR #634, opened 2023-05-08
Checking for status before adding units to browse list

myucekul:
Removing hidden units from browse list.


hardyoyo (approved):
obviously correct, 👍

PR #635, opened 2023-05-11
Added null check before flatten

myucekul:


hardyoyo (approved):
👍 looks good to me

hardyoyo (approved):
👍 looks good to me

PR #646, opened 2023-10-31
Use Code2000 for splash pages in SpashGen

hardyoyo:
For many years, DejaVuSans has been our font for our coverpages/splash pages. It's a reasonable choice, it is well-known for having good coverage of many unicode glyphs. But Code2000 has many more glyphs, and we need those glyphs for our coverpages.


everreau (approved):
seems simple enough 👍

PR #649, opened 2023-12-19
Script to enable unit id update

myucekul:
I have the change in dev using sql. Saving the script here. The process I used involved looking at the foreign constraint id for each and using that to drop. Do you recommend exploring the ruby way?

### DIFF
--- a/once_only/unit_id_cascade.sql
+++ b/once_only/unit_id_cascade.sql
@@ -1,3 +1,5 @@
+-- Update all tables with foreign key relations with unit id to cascade update to unable unit id changes
+
 ALTER TABLE category_stats DROP FOREIGN KEY category_stats_ibfk_1;
 ALTER TABLE category_stats ADD FOREIGN KEY (unit_id) REFERENCES units(id) ON UPDATE CASCADE;
 ALTER TABLE issues DROP FOREIGN KEY issues_ibfk_1;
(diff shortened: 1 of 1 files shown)


everreau (commented):
Looks OK to me.  Is there anyway to add a comment/documentation about this script's use and explain conceptually what all this is doing?

myucekul:
> Looks OK to me. Is there anyway to add a comment/documentation about this script's use and explain conceptually what all this is doing?

Done. 
I found another directory with ruby files with DB schema changes (https://github.com/eScholarship/jschol/tree/master/migrations). Do you think it is okay to do this one off change using sql commands or should I look into creating a migration file. I don't know what process was used to create the migration files. The last migration file was created 5 years ago.

bibliotechy:
@myucekul The biggest benefit to writing the change as a ruby migration is that it would be certain to be applied every time the DB is created, whether for a new instance for migration, or for a local instance for testing. 

Additionally, your script works on the DB as it is now, but may not work on the DB as it will be in the future, when other schema changes are made. That could impact the ability to run it in the future (for example of the name of the column or constraint gets changed).

With `Sequel`, migrations are generally not generated like they are in Django, but have to be written by hand. 

[The Sequel MIgration docs](https://sequel.jeremyevans.net/rdoc/files/doc/migration_rdoc.html) are pretty good for understanding how they are structured.

I think the command you will need is [drop_foreign_key](https://github.com/jeremyevans/sequel/blob/master/doc/schema_modification.rdoc#drop_foreign_key-), particularly the section that starts with "If you want to just drop the foreign key constraint without dropping the column, use an array."

If we don't expect to build the DB from scratch anywhere else, maybe this is not worth the effort, but mixing migrations and scripts like this certainly makes it harder to do so in the future.

myucekul:
Ruby sequel way is working well. I have cloud9 and RDS setup in pad-dev account. I plan to delete the DB there once I develop the script this week. I'll send another pull request for review soon. Thanks!

hardyoyo (approved):
👍

PR #650, opened 2024-01-11
Updated key constraints to cascade on update for unit id

myucekul:


bibliotechy (approved):
@myucekul This looks good. My only comment is that I'd like to see some inline comments linking out to the specific part of the Sequel docs that indicate how to drop constraints without dropping the foreign key itself. It's not obvious from the syntax, so a small comment could help future readers of this migration to understand what is happening.

myucekul:
> @myucekul This looks good. My only comment is that I'd like to see some inline comments linking out to the specific part of the Sequel docs that indicate how to drop constraints without dropping the foreign key itself. It's not obvious from the syntax, so a small comment could help future readers of this migration to understand what is happening.

Good point. That square bracket is super important.

migrations/0035_unitidcascade.rb
```diff
@@ -1,5 +1,6 @@
 Sequel.migration do
   up do
+    # update foreign key constraints to cascade on unit id update
```
hardyoyo, line 3: @bibliotechy's suggestion to add a link to the Sequel docs here is a good one, this link might be good enough?
https://github.com/jeremyevans/sequel/blob/master/doc/schema_modification.rdoc#add_foreign_key-

myucekul: agree

PR #651, opened 2024-01-22
Remove Codebuild configs, replace with GitHub workflow

hardyoyo:
- Use Reviewdog for lint checks (eslint and rubocop)


everreau (approved):
Looks OK to me.

PR #665, opened 2024-02-06
Remove all old Splashgen-related code, configs and dependencies

hardyoyo:
This code has moved to https://github.com/eScholarship/cdl-splashgen


hardyoyo:
Note that some Splash-page-related code remains here in Jschol, and will be difficult to refactor out, as this code is intertwined with the JSchol code: the splashGen.rb script, which is used by the Controller to generate splash pages for items as they are submitted, has two requireRelative lines that import Ruby libraries from Jschol.

myucekul (approved):
looks fine to me.

PR #675, opened 2024-02-21
implement in issue-level setting for showing or hiding workflow dates

everreau:
implement in issue-level setting for showing or hiding workflow dates


hardyoyo (approved):
Code looks fine, I haven't tested this, is it on dev? I'm approving just based on a code read.

PR #679, opened 2024-03-18
add default show pub dates setting to new issues

everreau:
add default show pub dates setting to new issues


hardyoyo (approved):
I'd love to see a short comment at the top of the method to give a brief synopsis for what's going on here, like line 811 below this new method. But I won't hold this up for that comment, I'd just like to see it at some point.

PR #687, opened 2024-05-29
Add dependency and AWS credential validation to deployVersion.sh and promoteVersion.sh scripts

hardyoyo:
Adds some dependency checks and an AWS credential validation to the deployVersion.sh and promoteVersion.sh scripts.


everreau (approved):
looks OK to me.

PR #690, opened 2024-06-04
PUBD-1313 Display HTML markup in requested fields

hardyoyo:
- Display HTML markup in issue title, issue description and issue cover image caption fields
- Disable truncation for issue cover image captions if these contain HTML markup

### DIFF
--- a/app/jsx/components/IssueComp.jsx
+++ b/app/jsx/components/IssueComp.jsx
@@ -45,11 +45,26 @@ class IssueComp extends React.Component {
       {p.cover &&
         <figure className="c-issue__thumbnail" ref={e => this.thumbnail = e}>
           <LazyImageComp src={"/cms-assets/"+p.cover.asset_id} alt="Issue cover" clickable={true}/>
-        {p.cover.caption &&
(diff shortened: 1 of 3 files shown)


hardyoyo:
a nice backgrounder of terser vs. other minifiers (like uglify): https://blog.logrocket.com/terser-vs-uglify-vs-babel-minify-comparing-javascript-minifiers/

The TLDR of commit [765ad32](https://github.com/eScholarship/jschol/pull/690/commits/765ad32841f3dfbf40c08472e68f2b92e0c0f871) is this:
uglify didn't zig with the zags, was still expecting md5 out of terser (whic uglify uses under the hood)... terser kept up with Node... so... just using terser alone gets us minification, where trying to use uglify gets us this error:
```
ERROR in app-bundle-1effd3cc50ea4d27ec9c.js from Terser
Error: error:0308010C:digital envelope routines::unsupported
    at new Hash (node:internal/crypto/hash:69:19)
    at Object.createHash (node:crypto:133:10)
    at /Users/hpotting/workspace/jschol/node_modules/terser-webpack-plugin/dist/index.js:217:37
```

everreau (approved):
looks OK to me

PR #691, opened 2024-06-05
PUBD-1327 Sort the series selectorList by the ordering column

hardyoyo:
Since the ordering column is present in the series selectorList props, let's use it to order the selectorList data. Fixes PUBD-1327.


hardyoyo:
deployed on dev and stg for testing, here's a path to a good test series on stg:

https://pub-jschol2-stg.escholarship.org/uc/cshe_150w_f100

remember to bring your key

everreau (approved):
👍

PR #692, opened 2024-06-06
Run the maybeSocks.rb tool before attempting to deploy.

hardyoyo:
- Should avoid situations where the proxied connection to S3 is
  dropped in the middle of a deploy


hardyoyo:
since I've tested this (works), and it's for a Lando dev environment, I'm just merging this

PR #693, opened 2024-07-05
Add devbox config

hardyoyo:
Adds configuration and changes a few things to facilitate using [Jetify Devbox](https://www.jetify.com/devbox) as a dev environment for JSchol. Also contains some changes to SCSS files to handle some deprecation warnings that we see during deploys and while running the dev environment.

### DIFF
--- a/.envrc.example
+++ b/.envrc.example
@@ -1,0 +1,87 @@
+# .envrc congifuration for jschol
+
+# WARNING: be sure to point at ONLY dev or stg for everything below... if you 
+# change from dev to stg or vice versa, *everything* needs to match. Recheck this,
+# you probably missed something if this isn't working the way you expect.
+
+### PUMA Configuration ###
+export PUMA_PORT=18880
+export PUMA_THREADS=5 # 5 is the default
+export PUMA_WORKERS=1 # a single worker is a good idea for a dev environment
+
+### ISO Configuration ###
+export ISO_PORT=4002
+
+# ### USE THE OJS DB ON THE DEV SERVER (uncomment to use, and also set USE_SOCKS_FOR_MYSQL=true, and pick the right SOCKS_TARGET)
+# export OJS_DB_HOST=ojs-host-dev
+# export OJS_DB_PORT='3306'
+# export OJS_DB_DATABASE=ojs
+# export OJS_DB_USERNAME=ojsuser
+# export OJS_DB_PASSWORD=ojspassword
+
+### USE THE OJS DB ON THE STG SERVER (uncomment to use, and also set USE_SOCKS_FOR_MYSQL=true, and pick the right SOCKS_TARGET)
+export OJS_DB_HOST=ojs-host-stg
+export OJS_DB_PORT='3306'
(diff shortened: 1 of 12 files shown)


setup.sh
```diff
@@ -1,22 +1,25 @@
...
 printf "== Installing local Ruby gems ==\n"
+gem install bundler
 bundle install --quiet --path=gems --binstubs
 
 #printf "\n== Uninstalling Ruby gems no longer used ==\n"
 bundle clean
 
 printf "\n== Installing node packages (used by gulp and iso via Node) ==\n"
-npm install npm@10.8.1
+npm install npm@10.2.4
```
everreau, line 18: Why is this going back to an older version of npm?

hardyoyo: Good catch, it should not, I will fix it

everreau (approved):
👍

app/scss/_utilities.scss
```diff
@@ -225,7 +225,7 @@ Background color with progressively-enhanced blur effect. Example:
 // Fluid text mixin courtesy of: https://css-tricks.com/snippets/css/fluid-typography/
 
 @function strip-unit($value) {
-  @return $value / ($value * 0 + 1);
+  @return calc(#{$value} / (#{$value} * 0 + 1));
```
myucekul, line 228: Please add comment on why calc is needed. Where is it defined?

setup.sh
```diff
@@ -14,9 +19,7 @@ npm install npm@10.8.1
 npm install
 npm install gulp-cli # shouldn't be necessary, but seems to be
 
-if [[ `/bin/hostname` == "*pub-submit*" ]]; then
```
myucekul, line 17: How do I know this submit specific code is no longer needed? setup.sh is not specific to dev box.

.envrc.example
```diff
@@ -0,0 +1,87 @@
+# .envrc congifuration for jschol
```
myucekul, line 1: This file looks like config/env.template except filled in. Where is it used?

app/scss/_utilities.scss
hardyoyo: This was one of the deprecation warnings. The previous version of the code uses division outside of a calc function, and will not be supported in a future version of the Sass library that we use for JSchol. [Calc is a built-in function in Sass](https://sass-lang.com/documentation/values/calculations/).

.envrc.example
hardyoyo: this is an example direnv configuration, used with direnv

setup.sh
hardyoyo: this is not how the splash page generator is built any more. See https://github.com/eScholarship/cdl-splashgen

app/scss/_utilities.scss
hardyoyo: here's a [message about the deprecation warnings](https://stackoverflow.com/questions/71262427/deprecation-warning-using-for-division-outside-of-calc-is-deprecated-and-wi), we had a wall of them every time we would build or deploy JSchol.

PR #694, opened 2024-07-10
Include stats for items redirected to an item

myucekul:
Included stats from items redirected to an item when calculating usage stats. Related card: https://trello.com/c/PQLnAKYJ


everreau (approved):
This makes sense to me though, I'm not familiar enough with this code to spot errors from code reading.  I'm happy to take a look at it when it's deployed to dev too.

PR #695, opened 2024-07-19
set charset to avoid encoding problem with longer Unicode chars

myucekul:
Need to set charset to correctly read the stored symbol from DB.


hardyoyo (approved):
looks good to me

PR #697, opened 2024-07-23
PUBD-1311 make default sort order more reasonable

hardyoyo:
This is a draft/work in progress, I'll mark it as so, and will let you know when it's ready to merge. As it is currently written, it uses the unit_id (any unit_id that starts with ucsd_jn_") to trigger a change from the default 'desc' sort order on series pages. This is a temporary fix, we'll need to add a "default_sort" field to unit properties, and then use it set the default sort for the unit.


hardyoyo:
replaced by #698

PR #698, opened 2024-07-24
For series use ordering_in_sect Pubd1311

myucekul:
The first change for series page is to use 'default' as default sorting. The second change is to utilize ordering_in_sect is available. The items with the order are placed first and then the remaining items.

### DIFF
--- a/app/unitPages.rb
+++ b/app/unitPages.rb
@@ -499,8 +499,9 @@ def getSeriesLandingPageData(unit, q)
   else
     children = parent ? $hierByAncestor[parent[0].ancestor_unit] : []
   end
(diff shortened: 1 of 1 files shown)


hardyoyo:
Aside from a few minor issues (including the failed linter issue from Rubocop) I think this ought to work. I'm going to give it s spin this morning, and will make a PR to this PR to include the suggested changes.

myucekul:
> Aside from a few minor issues (including the failed linter issue from Rubocop) I think this ought to work. I'm going to give it s spin this morning, and will make a PR to this PR to include the suggested changes.

Thank you. I have incorporated variable change and removed debug puts.

PR #700, opened 2024-07-24
Implemented series default sorting

myucekul:
Updated the series default logic. The last iteration was sorting within a page. The sorting by ordering_in_sect needs to happen for all items.


hardyoyo (approved):
👍 Looks good

PR #701, opened 2024-07-26
Limit special logic to only the series where pub_order is applied

myucekul:


app/unitPages.rb
```diff
@@ -582,6 +582,11 @@ def getIssueNumberingTitle(unit_id, volume, issue)
   return attrs['numbering'], attrs['title'], attrs['show_pub_dates']
 end
 
+
```
hardyoyo: If this empty method is still present in the deployed code, it should be removed.

PR #702, opened 2024-07-31
Restore date desc sort for non pub order case

myucekul:


hardyoyo (approved):
Looks good to me, and makes sense.

PR #708, opened 2024-08-16
Add cloudwatch config for memory stats

bibliotechy:
Adds a config file to send used memory percentage cloudwatch so we can track memory usage in the beanstalk instances.

Based on [AWS EB documentation](https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/customize-containers-cw.html).


myucekul (approved):
Looks good to me.

PR #709, opened 2024-08-16
Fixes for memory usage issues

myucekul:
1. Bringing back puma killer
2. Added GC.start in method related to pub_order change

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ backstop_data/
 bin/
 .bundle/
 cache/
-config/
 db_backup/
 dist
 gems/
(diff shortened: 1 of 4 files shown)


everreau:
Looks OK to me. Are we planning to leave the GC.start in long term?  Do we have data that it is helping? Or is it just a temporary shot in the dark?  At a minimum I think it should be commented.

bibliotechy (approved):
Looks fine to me. Let's just keep an eye on when these trigger in prod to make sure we are allowing enough memory for each process.

myucekul:
> Looks OK to me. Are we planning to leave the GC.start in long term? Do we have data that it is helping? Or is it just a temporary shot in the dark? At a minimum I think it should be commented.

Okay. I'll add comment. There is evidence that adding GC.start helps. Without is jschol running on t2.micro will die within 5 minutes just by user going to lbnl_rw series and reloading that repeatedly. With the change, it took almost an hour to get to that situation.

PR #765, opened 2025-02-27
Pubd 509 fix slug

nicktahani:
Fix for [CMS breaks with numbers in the URL slug (PUBD-509)](https://trello.com/c/tQ8avjdy). When creating a page with a number in the slug, it would return a 404. This updates the regex for valid slugs to also allow alphanumeric characters.

### DIFF
--- a/test/quick.rb
+++ b/test/quick.rb
@@ -211,4 +211,29 @@ class TestQuick < Test::Unit::TestCase
       puts "URL_PARAMS present, skipping test_for_broken_images"
     end
   end
+
+  def test_page_name_regex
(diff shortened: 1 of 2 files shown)


test/quick.rb
```diff
@@ -211,4 +211,29 @@ def test_for_broken_images
       puts "URL_PARAMS present, skipping test_for_broken_images"
     end
   end
+
+  def test_page_name_regex
```
bibliotechy, line 215: 🎉

bibliotechy (approved):

PR #789, opened 2025-04-22
Removed timestamp appending to PDF url

DevinSmithWork:
Removed the block that appends the timestamp. (I didn't see the displayPDF variable used anywhere else in server.rb).

I don't typically do jSchol deployments -- Can we do a quick sync up where you walk me through it? Thanks!


nicktahani:
Yeah, feel free to invite me whenever. These are the [deploy docs](https://github.com/cdlib/pad-sys-doc/wiki/jschol-deploy) I reference as an FYI.

DevinSmithWork:
Oh excellent! I'll see how far I get using the docs and get back to you if I run into anything disastrous

Issue #757, opened 2025-01-27
Update / Replace pdf.js


bibliotechy:
[Our fork of pdf.js](https://github.com/eScholarship/pdfjs-embed2) is extremely out of date (almost 9000 commits behind) and needs to either be updated or replaced.

PR #790, opened 2025-04-29
Replace PDF viewer 

nicktahani:
Fix for #757. Removed our [fork](https://github.com/eScholarship/pdfjs-embed2) of pdf.js and replaced with [react-pdf](https://github.com/wojtekmaj/react-pdf).


bibliotechy (approved):
Code looks good, let's this onto dev and get some program staff poking at it!

PR #797, opened 2025-05-08
use link with cover page for "download file"

nicktahani:


app/scss/_eschol_pdf_viewer.scss
```diff
@@ -65,3 +65,7 @@
 .c-pdfview__viewer {
   padding-top: 1em;
 }
+
```
bibliotechy, line 68: This fix is part of the URL issue? Just making sure it isn't included accidentally.

bibliotechy (approved):

app/scss/_eschol_pdf_viewer.scss
nicktahani: No, that was intentional. Not exactly related, just noticed the pdf was a little too cramped at the top.

Issue #798, opened 2025-05-13
Items in OA Policy series without abstracts should also have OA Policy banner


alainna:
### Why 

1. We want users who read OA Policy publications deposited to eScholarship to be informed that the work has been made available because of the UC's open access policies, and to give them the opportunity to tell us how being able to access the work benefits them.
2. This is already done for publications with abstracts; those without abstracts should have the same message. 

Compare: [Item with abstract: qt7jq9g523](https://escholarship.org/uc/item/7jq9g523) 

![Image](https://github.com/user-attachments/assets/ea0032ec-22c0-4b44-be7a-25342b123cbd)

[Item without abstract: qt42n2p8fp](https://escholarship.org/uc/item/42n2p8fp)

![Image](https://github.com/user-attachments/assets/f387a63e-817d-400e-b8b1-8f0596339ca0)

### What 

Enable the OA Policy disclaimer to display even when a publication does not have an abstract. 

> Many UC-authored scholarly publications are freely available on this site because of the UC's [open access policies](https://osc.universityofcalifornia.edu/open-access-at-uc/open-access-policy/). [Let us know how this access is important for you.](https://help.escholarship.org/support/tickets/new)

Mockup: 

![Image](https://github.com/user-attachments/assets/15edd400-5001-41da-bfbb-adeadf9fc26d)

nicktahani:
There's some publications that contain an abstract, but don't render the banner (e.g. https://escholarship.org/uc/item/2xg701jx). Looks like this happens because the unit id doesn't contain `_postprints`. Should this be updated as well?

PR #805, opened 2025-05-30
items in oa policy series without abstracts should also have oa policy banner

nicktahani:
Fix for #798. Banner will now render for all publications in OA Policy series, regardless of whether an abstract is present.


nicktahani:
Waiting for @justingonder's confirmation of correct series to proceed.

bibliotechy (approved):

Issue #811, opened 2025-06-24
jschol should properly display links in data availability statement


alainna:
### Why 

Authors can make good attempts to reference their data hosted on third party sites in their works, however their links can break because the subi submission form doesn't expect such. JSchol should help resolve some of these issues where possible, such as: 

1. Author has provided more than 1 data link, which is parsed as a single link, ex: https://escholarship.org/uc/item/03w2v8h7
2. Author provides plain text, which is parsed as a link, ex: https://escholarship.org/uc/item/1kc267nj

### What 

1. Author has provided more than 1 data link: jschol should interpret each as its own link, rather than one very long link 
2. Author provides plain text: if no HTTP or HTTPS, or TLDs referenced in the first part of the URL, jschol shoult interpret as plain text 

Dump of subi items data of objects with text or URLs in the data availability statement: https://drive.google.com/file/d/1NKytD1XUn8wOhBVgEYmZbz9aX1QOSZOC/view?usp=drive_link

suggestion re: display from the related  issue: 
![Image](https://github.com/user-attachments/assets/cc91b25e-10f7-4a83-9c5a-94ec4389cf1d)

Related: 

* https://github.com/cdlib/pub-subi/issues/42

PR #813, opened 2025-06-25
811 jschol should properly display links in data availability statement

nicktahani:
Fix for #811. Addresses the two issues listed in the ticket. A quick note:

- For data availability statements with a type of `"thirdParty"`, a corresponding contact property may be provided. This value can either be an entity name (e.g., `"California Air Resources Board, US EPA"`) or one or more URLs. The updated logic handles both cases: if the value contains URLs (starting with "http"), it renders each as an `<a>` link; otherwise, it renders the value as plain text inside a `<span>`.

### DIFF
--- a/app/scss/_pubinfo.scss
+++ b/app/scss/_pubinfo.scss
@@ -10,7 +10,6 @@
 
   @include bp(screen1) {
     display: grid;
-    grid-template-columns: 1fr auto;
     grid-gap: $spacing-sm $spacing-md;
 
     > *:not(.c-pubinfo__license) {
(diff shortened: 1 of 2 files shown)


app/scss/_pubinfo.scss
```diff
@@ -10,7 +10,6 @@
 
   @include bp(screen1) {
     display: grid;
-    grid-template-columns: 1fr auto;
```
myucekul, line 13: What is the reason for removal of this grid-template-columns?

myucekul (approved):
Looks fine to me.

app/scss/_pubinfo.scss
nicktahani: The mock-up in the ticket suggests displaying the links in a single column, but this styling arranges them in a single row using grid spacing. It also seemed to distort the text on smaller viewports:

Screenshot 2025-06-27 at 9 27 57 AM

PR #889, opened 2025-10-20
Dev env cleanup

everreau:
- Remove lando stuff including from docs
- Add a section in docs about current proxy requirements
- Remove references to bastion servers

@nicktahani there are currently 2 files that reference environment variables do we need both `.envrc.example` and `local.env.example`?  I think I only used the former for my setup but maybe the other is used for something else?


nicktahani:
I'm also only using `.envrc.example` for my setup. Looks like the other one is lando-specific – per `.lando.yml` it gets loaded here:

```yml
env_file:
  - defaults.env
  - local.env
```

so probably safe to remove.

PR #920, opened 2025-11-20
added OSTI ID to local labels on tab author comp

DevinSmithWork:
One-line change: Adding customized labels for OSTI IDs in the "Author & Article info" tab


DevinSmithWork:
Note: This change won't have any effect until OSTI IDs are in the eSchol DB, so it can be deployed asap

myucekul (approved):

PR #958, opened 2026-02-18
Disallow search for all bots

bibliotechy:


DevinSmithWork (approved):
Looks good. Should cut down on a bunch of churn

Issue #903, opened 2025-11-10
Finish jSchol upgrade to vite


bibliotechy:
@nicktahani Was almost done migrating the jSchol app to a new build tool, vite, that will allow us to add update dependencies, add tests, etc. 

We shoudl complete this work to unblock some future development.

Issue #769, opened 2025-03-07
Upgrade bundler


nicktahani:
Our current version of Webpack is outdated and restricts our ability to fully leverage newer libraries and features. We should consider exploring alternative solutions that provide better support for modern development tools and workflows.

Some options to explore:

1. Migrate to Vite (preferred)
2. Upgrade webpack v4 -> v5 ([guide](https://webpack.js.org/migrate/5/))

PR #969, opened 2026-02-26
webpack --> vite migration

nicktahani:
Addresses #769 and #903. Replaced Webpack + Gulp with Vite. A few things to note:

- Webpack and Gulp configs were complex, hard to maintain, and pulling in a large number of transitive dependencies
- Vite provides faster dev builds via native ESM, built-in HMR, and simpler config
- Removes Babel in favor of Vite's native JSX/ESM transform
- Source maps are only generated in development mode (intentional for bundle size)

**Added:**

- `vite.config.js`: single config for dev and prod (with SSR support)
- `isomorphic-entry.jsx` and `isomorphic-server.js`: new SSR entry points replacing `app/isomorphic.jsx`
- `main.jsx`: client entry point for Vite

**Updated:** 

- `package.json`: new scripts (`npm run dev` replaces `gulp` command to spin up app)
- `server.rb`: updated Vite manifest parsing and response handling
- `app/scss/*.scss`: updated Sass color functions to address deprecation warnings
- `LazyImageComp.jsx`: converted to functional component; replaced lozad with native `loading="lazy"`
- `README`: updated setup instructions (`./gulp` --> `npm run dev`)
- `deployVersion.sh`, `setup.sh`, `make-dev-symlinks.sh`: updated for Vite and removed Webpack+Gulp references 

**Known issues:**

- Sass deprecation warnings are currently silenced pending upstream fixes

**EDIT (3/4):**

- better SSR handling in the local dev environment: rather than bundling, we can use vite's `ssrLoadModule()` to load the entry point at request-time (for deployed versions we still use the pre-built bundle)
- fixed a silent React hydration bug where the page data was never embedded in the HTML response. without it, the client ignored the server-rendered markup and fell back to a full re-render, making the SSR work effectively a no-op. this was a pre-existing regression, not introduced by this PR

### DIFF
--- a/app/jsx/components/LazyImageComp.jsx
+++ b/app/jsx/components/LazyImageComp.jsx
@@ -1,8 +1,7 @@
 // ##### Lazy Image Component ##### //
 
-// NOTE:
-// chrome is very efficient for a small quantity of images, so loading='lazy' doesn't always work
-// however, using the native attribute eliminates the need for external dependencies 
+// NOTE: loading='lazy' may not always defer small sets of images (browsers apply their own)
+// viewport thresholds may load images eagerly if they're close enough
 
 import React, { useRef, useEffect } from 'react'
 


bibliotechy:
❤️❤️❤️ Screenshot 2026-03-09 at 2 03 19 PM

PR #974, opened 2026-03-05
Updating cover images for homepage

justingonder:


bibliotechy:
One suggestion I'm going to make it that we should convert png files to webp format. For example, homecover_sfnews.png is 108KB, but the webp I created locally is only 12KB, but the difference between them is undetectable. 

I was able to create webp with imagemagick on my laptop. It can also be done in photoshop. 

PNG
homecover_sfews

webp
homecover_sfews

bibliotechy (approved):

PR #1008, opened 2026-04-15
increase cache length

everreau:
increase the cache control header for static files and pdfs from 1 hour to 1 week


bibliotechy (approved):

everreau:
deployed to dev.  Looks OK to me.

PR #1038, opened 2026-05-15
Changes for Ruby3.4 version of EB

myucekul:
EB name change and updated gem lock file.


nicktahani:
This doesn't seem to work in the local devbox environment. I updated the ruby version in [devbox.json](https://github.com/eScholarship/jschol/blob/master/devbox.json), but it errors on the `bigdecimal` gem install. Investigating.

nicktahani (approved):

PR #1041, opened 2026-05-19
update ruby version to 3.4.8 in devbox config

nicktahani:
The issue mentioned [here](https://github.com/eScholarship/jschol/pull/1038#issuecomment-4480446371) resulted from my local devbox version being too old. Discussed in this [thread](https://github.com/jetify-com/devbox/issues/2502).


myucekul (approved):
Looks good. Thanks Nick!

Issue #951, opened 2026-01-29
jschol Beanstalk needs upgrade to Ruby 3.3 by July 2026


myucekul:
Related message:
[Action Required] AWS Elastic Beanstalk Ruby 3.2 AL2023 platforms retire July 31, 2026 [AWS Account: 777968769372] [US-WEST-2]
View details in service console
Hello,
[AWS Health may periodically trigger reminder notifications about this communication if resources remain unresolved.]
AWS Elastic Beanstalk is informing you that platform branches [1] based on Ruby 3.2 on Amazon Linux 2023 will retire on July 31, 2026. This is to ensure that your customer's Elastic Beanstalk environments are aligned with the most current support offered by AWS.
* What does retire mean, and what will happen after July 31, 2026?
Starting on July 31, 2026, retired platform branches will no longer be available for new environments on Elastic Beanstalk. Customers can continue to operate existing environments running on retired platform branches, but these branches will no longer be managed and supported by Elastic Beanstalk. For more information, see the Elastic Beanstalk platform support policy [2] in the AWS Elastic Beanstalk Developer Guide.
* What do I need to do now?
You have until July 31, 2026, to upgrade your Elastic Beanstalk environments configured on platform branches running on Ruby 3.2 on Amazon Linux 2023. We strongly recommend that you start planning your migration to one of the following Ruby platforms, which are current and fully supported:
- Ruby 3.3 AL2023
- Ruby 3.4 AL2023
Use the Elastic Beanstalk console or the EB CLI to update your environment's platform version, see "Updating your Elastic Beanstalk environment's platform version" [3] in the AWS Elastic Beanstalk Developer Guide.

myucekul:
Create a new EB instance for jschol dev

myucekul:
Create jschol dev EB with Ruby 3.4 stack.  Updated the dev url to point to the new EB.

myucekul:
Login results in error. Rest of the site is working. 
May 28 16:58:04 ip-10-62-9-197 puma[258527]: 2026-05-28 16:58:04 - Sequel::DatabaseError - Mysql2::Error::ConnectionError: Server has gone away (Sequel::DatabaseError):
May 28 16:58:04 ip-10-62-9-197 puma[258527]: #011/var/app/current/vendor/bundle/ruby/3.4.0/gems/mysql2-0.5.7/lib/mysql2/client.rb:146:in 'Mysql2::Client#_query'
[...]

PR #1044, opened 2026-05-28
Added keep alive for OJS DB

myucekul:
Added reconnect to config as well as test flag to validate connection. 
Call stack for the OJS DB failure after instance idle is in the issue:
https://github.com/eScholarship/jschol/issues/951

### DIFF
--- a/app/server.rb
+++ b/app/server.rb
@@ -82,10 +82,7 @@ def ensureConnect(envPrefix)
                "charset"  => 'utf8mb4',
                "reconnect"      => true,
                "max_connections"=> 10,
-               "pool_timeout"   => 5
-
-
-  }
+               "pool_timeout"   => 5}
   if ENV['USE_SOCKS_FOR_MYSQL'].to_s.downcase == 'true'
     if TCPSocket::socks_port
       SocksMysql.new(dbConfig)
--- a/app/server.rb
+++ b/app/server.rb
@@ -79,16 +79,17 @@ def ensureConnect(envPrefix)
                "database" => getEnv("#{envPrefix}_DATABASE"),
                "username" => getEnv("#{envPrefix}_USERNAME"),
                "password" => getEnv("#{envPrefix}_PASSWORD"),
-               "charset"  => 'utf8mb4',
-               "reconnect"      => true,
-               "max_connections"=> 10,
-               "pool_timeout"   => 5}
+               "charset"  => 'utf8mb4'}
+
+  if envPrefix == "OJS_DB"
+      dbConfig["test"] = true
+  end
   if ENV['USE_SOCKS_FOR_MYSQL'].to_s.downcase == 'true'
     if TCPSocket::socks_port
       SocksMysql.new(dbConfig)
     end
   end
-  db = Sequel.connect(dbConfig.merge(test: true))
+  db = Sequel.connect(dbConfig)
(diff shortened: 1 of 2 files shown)


app/server.rb
```diff
@@ -79,13 +79,19 @@ def ensureConnect(envPrefix)
                "database" => getEnv("#{envPrefix}_DATABASE"),
                "username" => getEnv("#{envPrefix}_USERNAME"),
                "password" => getEnv("#{envPrefix}_PASSWORD"),
-               "charset"  => 'utf8mb4' }
+               "charset"  => 'utf8mb4',
+               "reconnect"      => true,
+               "max_connections"=> 10,
```
martinhaye: Your commit didn't mention setting max_connections and pool_timeout - are these values based on something?

Also, I wondered if this logic is used for both the connections - jschol and ojs. If yes for jschol, this could have an impact (positive or negative, I'm not sure).

app/server.rb
```diff
@@ -79,13 +79,19 @@ def ensureConnect(envPrefix)
...
+
+
+  }
   if ENV['USE_SOCKS_FOR_MYSQL'].to_s.downcase == 'true'
     if TCPSocket::socks_port
       SocksMysql.new(dbConfig)
     end
   end
-  db = Sequel.connect(dbConfig)
+  db = Sequel.connect(dbConfig.merge(test: true))
```
martinhaye: Haven't seen the 'merge' thing before - what is that?

martinhaye (commented):
I put a couple questions inline.

app/server.rb
myucekul: The function is used for both jschol and OJS. I can make the change specific to OJS once I have confirmation that the login problem after long idle time is solved. For OJS connection, I don't see background tasks/threads using it. I may not need max_connection setting. 

test:true is to validate connection.  From copilot:
Even if you enable reconnect: true, Sequel only reconnects on the next query, not in the background.
If your app holds a connection for a long time (e.g., long-running threads), you still need test: true so Sequel validates the connection before handing it out.

app/server.rb
martinhaye: Seems like `"test" => true` could be added directly to dbConfig, rather than this little confusing merge thing.

myucekul: Agree. I'll add a check for OJS and then add reconnect and test to the config. That will cleaner and more focused change.

myucekul: Interesting - I see warning in puma log:
WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.

Copilot says:
The correct modern fix: use test: true instead of reconnect: true

nicktahani:
@myucekul log-in worked for me locally, I do get these warnings:

```
[puma] [0] Connecting to eschol DB.    
[puma] WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
[puma] [0] Connecting to OJS DB.       
[puma] WARNING: MYSQL_OPT_RECONNECT is deprecated and will be removed in a future version.
```

app/server.rb
myucekul: @nicktahani - thanks for sharing your observation. Let me update PR and apply the change in dev. I'll let the instance be idle for rest of the day and test tomorrow morning.

martinhaye (approved):
Looks good to me. Hoping it works.

PR #1055, opened 2026-06-08
WIP: Enable caching for item pages

myucekul:
Item page caching is working in stage.


DevinSmithWork (approved):
LGTM -- did you end up replacing the unescaping with a different strategy?

myucekul:
> LGTM -- did you end up replacing the unescaping with a different strategy?

No, it is still there. Just moved to line 367. I am using the unescaped value for /uc/item check.

